PR #14, opened 2026-05-18
Fix entity-resurrection + spell DSL crashes; add property-testing suite

nnunley:
Hi! Long-time admirer of let-go and xsofy — picked this up from a crash dump I hit (player on fire stepping onto an item) and the work grew a bit. Splitting it into a single PR here; happy to break it up or drop pieces if anything's unwelcome.

### Crash fixes (`da85543`)
Three related bugs where dead-or-removed entities were being resurrected as partial shells:

- **`pickup-item`** was calling `update-in [:entities owner-id :inventory] conj item-id` without checking whether `owner-id` still existed. If the player died on the same turn they tried to pick something up (burning + step-on-item), `update-in` would happily recreate `{:inventory [item-id]}` under `:player`. Now bails out when owner is missing.
- **`tick-entity-statuses`** had the same shape: each per-status reduce step assumed the entity was still there, but `:burning` could kill on tick and the next status would resurrect them. Each step now checks presence.
- **`apply-status` / `remove-status`** were the same pattern at the API level — now defensive no-ops on missing entities.
- **`r-spread`** in the spell DSL threw on an empty world. let-go's lazy `sort-by` composed with `filter` on an empty input nil-pointers (and `(set (filter some? (map :pos empty)))` produces `#{nil}` rather than `#{}`). Fixed with an empty-cursors early return and forcing eager `vec` at the boundaries. Comments in the code call this out as a let-go quirk in case it bites again.

### Player-facing display fix (`06d0bd5`)
The rune inscription UI was leaking the parser's internal error map when a token sequence failed to parse — e.g. inscribing `:add` with no args showed `(magic-> kryl {:reason :bad-argument :rune :add :args [nil nil] :error true})` in the spell display.

`format-sexp-step` now detects the error-map shape and renders 2-3 random glyphs from the existing rune-glyphs set. The glyphs shift on each render — the runes appear to be failing to bind into a stable form. Felt more in keeping with the discovery aesthetic than English error text.

### Property-testing library (`1ff48d5`, `48b339b`, `9b127d7`, `2b8520e`, `f6c53ab` — last one is the collision-resolver, ignore here)
Added `xsofy.check`, a QuickCheck-style property testing library written to fit let-go (around its eager `map`, lazy-seq composition gotchas, and missing `reduce-kv`). Splittable LCG-based PRNG, rose-tree-based integrated shrinking, generators with bind/fmap/`gen-let`, `defprop` for terse property definitions.

`xsofy.check-gen` is the domain layer — generators for positions, statuses, players, minimal/feature/rich worlds, plus a `world-invariants?` predicate that's the single source of truth for what a structurally valid world looks like (8 invariants — see commit message for the list). Worth noting: creature-on-creature positional collision is *not* in the invariant list — that's a runtime policy decision, not a structural one, and treating it as an invariant made generators useless for fuzz testing.

Tests are in `xsofy/test/`:
- `check_test.lg` — 53 unit tests for the library itself (generators, RNG, shrinking, defprop, frequency distribution)
- `invariant_test.lg` — pins each of the 8 invariants and verifies generators don't produce invalid worlds
- `spell_prop.lg` — fuzz tests the spell DSL with both raw token sequences and well-formed programs (the most crash-prone surface); includes the targeted regressions for the bugs above
- `world_prop.lg` — movement, fire simulation, status ticks
- `combat_prop.lg` — combat, push, save/restore, equipment, AI

Runner: `lg xsofy/test/run.lg`. Current run: 105 tests, 136 assertions, 0 fail, 0 error.

## Why one PR

The crash fixes alone are small, but every one of them was found (or, after the fact, pinned) by a property test. Shipping the fixes without the suite means losing the regression coverage. If you'd rather take the crash fixes first and the property infra in a follow-up, I'm happy to split — just say the word.

## Won't break anything I can see

No public-API changes beyond the new `xsofy.check` and `xsofy.check-gen` namespaces. `world-invariants?` is exported but nothing in the existing code uses it. The `r-spread`, status, and entity changes are all narrow guards — nothing existing should hit the new code paths.

### DIFF
--- a/main.lg
+++ b/main.lg
@@ -476,9 +476,11 @@
     (game-loop (new-game))
     (catch e
            (dbg/log "FATAL ERROR:" e)
+           (dbg/dump-crash! e)
       (shutdown-terminal)
       (println "FATAL ERROR:")
-      (println e))
+      (println e)
+      (println "World state dumped to crash-dump.edn"))
     (finally
       (shutdown-terminal))))
 
--- a/xsofy/check.lg
+++ b/xsofy/check.lg
@@ -1,0 +1,514 @@
+(ns xsofy.check
+  (:require [test]))
+
+;; ============================================================================
+;; RNG — splittable linear congruential generator
+;; State is a 2-vec [state gamma] so we can split into independent streams.
+;; ============================================================================
+
+;; LCG constants from Numerical Recipes
+(def ^:private rng-mul 1664525)
+(def ^:private rng-add 1013904223)
+(def ^:private rng-mod 4294967296)   ;; 2^32
+
+(defn rng
+  "Create an RNG state from a seed integer."
+  [seed]
+  ;; gamma is the increment; odd gammas keep the LCG period full
+  [(mod (or seed 0) rng-mod) (bit-or 1 rng-add)])
+
+(defn rng-next
+  "Returns [n rng'] — n is a non-negative integer < 2^31, rng' is the next state."
+  [r]
+  (let [[s g] r
+        s' (mod (+ (* s rng-mul) g) rng-mod)
+        ;; take the high bits — LCG low bits have short periods
+        n (mod (quot s' 2) 2147483648)]   ;; < 2^31
+    [n [s' g]]))
+
+(defn rng-split
+  "Split an RNG into two independent generators with distinct gammas."
+  [r]
+  ;; Advance once to derive a fresh state, then build a sibling with a
+  ;; different (odd) gamma. The two children diverge from the next step.
+  (let [[n r'] (rng-next r)
+        [s g] r'
+        sibling-gamma (bit-or 1 (+ g n 2654435769))]   ;; golden-ratio mix
+    [r' [s sibling-gamma]]))
+
+;; ============================================================================
+;; Rose tree — {:value v :shrinks (lazy-seq of rose trees)}
+;; ============================================================================
+
+(defn pure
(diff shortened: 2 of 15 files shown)


nooga:
Oh wow! This is amazing @nnunley 🤩 I'm floored!

I've been playing very fast and loose with this project - it was supposed to be a fun demo for let-go based on a vague game idea. Your approach makes it much more manageable. Esp. Fire & effects + entity lifecycle were always super shaky.

Completely happy with one big PR - I'll try to review tonight and merge unless anything serious pops up.

Btw. I think that `check` could belong in let-go itself - we could bundle it there and grow it into a proper check-style library. Would that be interesting to you? For the time being I'll merge it here.

nnunley:
> Oh wow! This is amazing @nnunley 🤩 I'm floored!
> 
> I've been playing very fast and loose with this project - it was supposed to be a fun demo for let-go based on a vague game idea. Your approach makes it much more manageable. Esp. Fire & effects + entity lifecycle were always super shaky.
> 
> Completely happy with one big PR - I'll try to review tonight and merge unless anything serious pops up.
> 
> Btw. I think that `check` could belong in let-go itself - we could bundle it there and grow it into a proper check-style library. Would that be interesting to you? For the time being I'll merge it here.

I'm entirely willing to port check over to let-go. I just wanted to reduce my patch surface.

nooga (approved):

Issue #17, opened 2026-05-20
start new game after death stats


simon1tan1:
see title. Reaching the "program terminated", refreshing doesn't load a new game. I thought you had fixed it on the 5/16 - perhaps not push to production yet?

nooga:
I also thought this was fixed but apparently not. I'll look into this 👍

nooga:
Fixed, closing

PR #23, opened 2026-05-22
Fix web restart and floor fire state

### LINKED ISSUE #21: fire tiles remain after stair use
Web version.

when going up stairs to and then coming back down, the fire tiles continue showing. I can walk though without getting burned.

nooga:
## Summary
- Preserve fire TTL when saving/restoring floors so returning to a floor does not leave harmless visual fire tiles behind.
- In web builds, prevent Escape from exiting to the terminal/program-exited screen by confirming restart on the death screen.
- Update HUD/death-screen prompts to say restart in web mode and keep native quit wording outside web mode.

## Related issues
- Fixes #21
- Related to #17

## Tests
- lg xsofy/test/run.lg
- lg -b /tmp/xsofy-check main.lg

### DIFF
--- a/main.lg
+++ b/main.lg
@@ -376,14 +376,29 @@
                      (recur w w (term/size))))))
 
          :death
-         (loop [scroll 0]
-           (render/render-death-screen world scroll)
-           (let [k (term/read-key)]
-             (cond
-               (= k "\u001b") (assoc world :running false)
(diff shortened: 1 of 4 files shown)


nnunley (commented):
So, late to the review process, but... I think it might help, more broadly, if you have a set of protocols for rendering and input handling so you don't have to litter the code with conditional checks for the wasm environment vs terminal.

PR #29, opened 2026-05-23
Fix Pages COI patcher splice

nooga:
## Summary
- avoid raw index slicing when replacing the generated COI bootstrap block
- assert the WASM bootstrap declaration survives patching

## Verification
- lg -w dist main.lg
- lg tools/patch_wasm_coi.lg
- node -e "const fs=require('fs'); const html=fs.readFileSync('dist/index.html','utf8'); const m=[...html.matchAll(/<script(?: [^>]*)?>([\\s\\S]*?)<\\/script>/g)]; new Function(m[m.length-1][1]); console.log('inline script parses');"

### DIFF
--- a/tools/patch_wasm_coi.lg
+++ b/tools/patch_wasm_coi.lg
@@ -14,6 +14,9 @@
 (def wasm-marker
   "// --- Inline wasm_exec.js and WASM data ---")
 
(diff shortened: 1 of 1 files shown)


nnunley (approved):
lgtm.

PR #35, opened 2026-05-27
Implement conjure, stub friend, and fix loot backfires (#32)

### LINKED ISSUE #32: Conjure friend backfires, conjure not implemented
Per the spell-dsl, conjure without apply should summon a random entity and friend should modify that conjuration so the entity is friendly. Instead it backfires.
It's looking like conjure might not be implemented yet, so that probably is the source of the error causing the backfire.

mparrett:
Closes #32.

This is **one possible path** — primitives implemented to remove the backfire, with several deliberate design trade-offs marked below. Happy to redirect any of them based on your read.

## Bugs

1. `:conjure` and `:friend` were in `all-runes` and documented in `docs/spell-dsl.md` but never landed in `rune-primitives`. Inscribing either triggered `:unknown-rune`, which `apply-spell-result` converted into the on-hit backfire that zapped the caster.
2. The `:useful` loot tier emitted `(apply <effect> N)` for every effect in `runic-effects`, including `:push` and `:arc` — neither accept a numeric arg. `valid-rune-args?` rejected them with `:bad-argument` and the on-hit fired as a backfire. That's the Thundering Flail the reporter described.

## Implementation

**`r-conjure`** — picks a random species from a hard-coded `conjurable-species` list, finds an adjacent walkable + unoccupied cell, spawns via the existing `ent/spawn`. Stores the new id on the ctx as `:last-conjured-id`.

**`r-friend`** — reads `:last-conjured-id` and sets `:friendly true` on that entity. **Marker only — AI doesn't honor it yet** (see open questions).

**Soft-fizzle path** — when `r-conjure` finds no adjacent walkable cell, it sets `:fizzle` on the ctx instead of returning unchanged or throwing. `eval-spell` propagates `:fizzle` alongside `:error` in the result map; `apply-spell-result` formats it with a parallel rune-on-source message — quieter than backfire, no damage:

> `Rune <glyph> on the <weapon> hums but nothing answers.`

**Loot fix** — `runic-effects` gains `:takes-arg`. `:useful`-tier shape 1 gates `(apply <effect> N)` on it; when the picked effect isn't parametric, the slot falls back to shape 0 (bare effect), keeping the effect's flavor name.

## Open questions / liberties taken

I'd value direction on any of these:

1. **`r-friend` is a marker, not a behavior.** The AI in `xsofy.ai` has no faction concept — every non-player entity hunts the player. Making `:friendly true` actually friendly needs an AI refactor (target acquisition picks nearest enemy; player combat excludes friendlies; etc.). I deliberately punted that, so today `friend` shows up in the inscription and parses cleanly but doesn't visibly change creature behavior. Alternatives: implement the AI work in this PR; drop `:friend` from the rune vocabulary until faction lands; leave as-is with a known-gap note.

2. **`conjurable-species` is hard-coded and ungated.** Currently any creature in the bestiary is fair game — including bosses (dragon, troll, wraith, ogre). Could be depth-gated, weighted by tier, or sourced from a different list. Opinions welcome.

3. **No support for `(apply conjure <entity-type>)`.** The spec mentions entity-type runes (`docs/spell-dsl.md:58`) but the rune vocabulary doesn't currently include creature keywords, so there's no inscription path. I kept the implementation single-arity (random species). If entity-type runes are intended to land separately, this primitive can extend to accept them later without breaking changes.

4. **`:fizzle` as a new result key, parallel to `:error`.** I added new vocabulary rather than overloading `:error` with a `:reason :fizzle` and special-casing in `apply-spell-result`. Reasoning: fizzle is semantically distinct (soft failure, no damage) and the new key keeps the existing backfire path untouched. Open to merging into `:error` if you'd rather keep result-map shape minimal.

5. **Fizzle message wording** — `"Rune X on the SOURCE-NAME hums but nothing answers."` Tried to match the existing terse-atmospheric register (`"The runestone crumbles to dust."`, `"The wraith dissolves into mist."`). Pitch alternatives freely.

6. **Loot generator: `:takes-arg` field vs separate set.** Added a per-effect field so each entry is self-describing. Alternative shapes: a `parametric-effects` set, or query the existing `rune-arg-kinds` map at pick time (single source of truth). The field is local and lightweight; the cross-reference would be DRY-er.

7. **Shape-1 fallback when non-parametric.** When the rolled effect can't take a numeric arg, useful-tier shape 1 degrades to shape 0 (bare effect, same name). Alternative: re-roll the shape, or pick a different effect. Current behavior keeps the effect's flavor name visible.

## Tests

Five new tests in `spell_prop.lg`:

- `conjure-summons-an-adjacent-creature` — happy path; entity count goes up, position is one of 8 neighbors
- `conjure-then-friend-flags-the-new-entity` — chained primitives set `:friendly`
- `friend-without-prior-conjure-is-a-quiet-noop` — `:friend` alone doesn't error
- `generated-runic-programs-pass-spell-validation` — 400-iter sweep (4 tiers × 100) against `valid-rune-args?`; catches the `(apply push N)` / `(apply arc N)` family
- `conjure-without-room-surfaces-a-fizzle-not-an-error` — verifies `:fizzle` is set without `:error`

Stash-and-rerun verified each fix's tests catch the production bug.

```
Finished running tests. Tests: 167 Pass: 1359 Fail: 0 Error: 0
```

### DIFF
--- a/xsofy/items.lg
+++ b/xsofy/items.lg
@@ -292,11 +292,17 @@
 ;; --- Procedural Runics ---
 
 ;; Effect pools with weights
+;; :takes-arg gates the (apply <effect> N) inscription shape — only
+;; effects that accept a numeric arg can use it. Without this guard,
+;; the loot generator emitted programs like (apply arc 3) on the
+;; Thundering Flail, which the spell parser rejected with :bad-argument
+;; and apply-spell-result converted into a backfire zap on every hit
+;; (nooga/xsofy#32).
 (def runic-effects
-  [{:effect :fire  :weight 20  :name "flaming"}
-   {:effect :ice   :weight 20  :name "freezing"}
-   {:effect :push  :weight 15  :name "repulsing"}
-   {:effect :arc   :weight 5   :name "thundering"}])
+  [{:effect :fire  :weight 20  :name "flaming"     :takes-arg true}
+   {:effect :ice   :weight 20  :name "freezing"    :takes-arg true}
+   {:effect :push  :weight 15  :name "repulsing"   :takes-arg false}
+   {:effect :arc   :weight 5   :name "thundering"  :takes-arg false}])
 
 (def runic-self-effects
   [{:effect :heal    :weight 15  :name "vampiric"}
@@ -331,8 +337,17 @@
         (case shape
           0 [{:runes [(:effect eff)]
               :name (:name eff)} world]
-          1 [{:runes [:apply (:effect eff) param]
-              :name (:name eff)} world]
+          ;; Shape 1 requires a parametric effect — fall back to shape 0
+          ;; when the picked effect doesn't accept a numeric arg. Without
+          ;; this guard the loot generator emits programs like
+          ;; (apply arc 3) on the Thundering Flail, which the spell parser
+          ;; rejects with :bad-argument and apply-spell-result converts
+          ;; into a backfire zap on every hit (nooga/xsofy#32).
+          1 (if (:takes-arg eff)
+              [{:runes [:apply (:effect eff) param]
+                :name (:name eff)} world]
+              [{:runes [(:effect eff)]
+                :name (:name eff)} world])
           2 (let [[eff2 world] (pick-effect
                                 world
                                 (vec (filter #(not= (:effect %) (:effect eff))
--- a/xsofy/spell.lg
+++ b/xsofy/spell.lg
@@ -275,11 +275,30 @@
               (:cursors ctx))]
       (assoc ctx :world w))))
 
-;; Creatures the player can conjure. Excludes :player. Future versions may
-;; gate by depth or weight; for now any non-player species is fair game.
+;; Creatures the player can conjure, with pick weights by bestiary tier.
+;; The bestiary's "DEPTH N-M" section comments map species to tiers, and we
+;; bias inversely: tier-1 creatures (rat-tier) appear ~6× more often than
+;; tier-5 (wraith). The combined boss-tier probability is ~8% — possible
+;; "OP run" outcomes happen but aren't the median experience.
+;; Tuned per nooga/xsofy#32 review feedback (Oakwhisper).
 (def conjurable-species
-  [:rat :goblin :kobold :archer :spider :slime
-   :goblin-shaman :fire-imp :ogre :troll :wraith])
+  [[:rat 6] [:goblin 6] [:kobold 6]              ;; tier 1 — depth 1-2
+   [:archer 4] [:spider 4] [:slime 4]            ;; tier 2 — depth 2-3
+   [:goblin-shaman 2] [:fire-imp 2]              ;; tier 3 — depth 3-4
+   [:ogre 1] [:troll 1]                          ;; tier 4 — depth 4-5
+   [:wraith 1]])                                 ;; tier 5 — depth 5-6
+
+(defn- weighted-pick
+  "Pick a key from a vec of [k weight] pairs, biased by weight."
+  [pairs]
+  (let [total (reduce + (map second pairs))
+        roll (rand-int total)]
+    (loop [pairs pairs acc 0]
+      (let [[k w] (first pairs)
+            acc' (+ acc w)]
+        (if (> acc' roll)
+          k
+          (recur (rest pairs) acc'))))))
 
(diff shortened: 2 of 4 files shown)


nooga:
I think it looks good but need to take some time to think about this one before merging. Will check back tomorrow :)

Oakwhisper:
I think some level of weighting for conjure random spawns should be taken into account. Otherwise with the effectively unlimited amount of entities that could be summoned, getting an overpowered ally could be too great. Making it impossible to get boss monsters however I think is no fun. The random nature of the game should be able to bestow a player with an OP run, but high probability is a bit much

Implementing conjure should also probably introduce the friend/enemy concept to the AI otherwise unsuspecting players will just get themselves killed while trying to properly use the feature

mparrett:
Thanks @Oakwhisper, both points make sense to me. I pushed `71c26b8` on top of the existing branch so we can assess; this way it's easy to inspect or revert.

**1. Weighting.** Switched `conjurable-species` from a flat vec to `[species weight]` pairs biased by the bestiary's existing depth-tier comments. A small `weighted-pick` helper draws against the total weight. Distribution:

| Tier | Species | Weight ea. | Combined % |
|---|---|---|---|
| 1 (depth 1-2) | rat, goblin, kobold | 6 | 48.6% |
| 2 (depth 2-3) | archer, spider, slime | 4 | 32.4% |
| 3 (depth 3-4) | goblin-shaman, fire-imp | 2 | 10.8% |
| 4 (depth 4-5) | ogre, troll | 1 | 5.4% |
| 5 (depth 5-6) | wraith | 1 | 2.7% |

These numbers are a first pass. Boss-tier combined is ~8% and this seems reasonable for a possible OP run, not median. Could tune it lower to start, since dialing it might require play testing and feedback. I think the key balance question for the first cut is whether to err on the side of a lower chance of OP run or higher chance. All of this will be subject to a later balancing pass, so we just need to converge on a reasonable starting point we can live with.

**2. `r-friend` is now a stub.** Marker without a consumer was the original cut. I think you're right that it's strictly worse than not promising friendliness at all. The rune is still in `rune-primitives` so inscriptions don't backfire (the original #32 bug), but the conjured creature is no longer flagged `:friendly`. When faction-aware AI is introduced, flipping `r-friend` back to set the marker is one line. The renamed test `conjure-then-friend-stub-is-noop-until-faction-ai` documents the choice and gap.

**On the faction follow-up:** as this PR has enough complexity and design decision load, I recommend we tackle factions in a separate PR. I think this would include: target acquisition in `xsofy.ai`, player-move combat exclusion, and probably a render glyph distinction? Probably best to file as its own issue to sort out the design surface (cross-floor persistence of friendly summons? what happens when a friendly creature kills a hostile? can it loot? who gets the gold?) to discuss before any coding takes place.

Thoughts/concerns about on the overall shape here — weighted approach (and weights) — as a first pass for conjure? An alternative/hybrid could be depth-aware curve that's nonzero at or below your current depth? It's possible @nooga has something entirely different in mind, like cost in HP/turns scaling with tier. Happy to take a stab at tuning the weights or a different implementation.

nooga (dismissed):

nooga:
@mparrett I've commented in #37 regarding this one. Please observe recently merged #39 and determine if it affects this work. If all looks good after rebase I think we can land this. Thanks!

mparrett:
Thanks @nooga — appreciate your comments and will report on how the rebase goes. I also have a draft ready for #37 that can land after this one.

mparrett:
@nooga Rebase wasn't trivial nor super hairy. Term/browser smoke test was ok. Appreciate if you could glance at 4ffef07a59b9c94e5f116fb3682b2a483f8c5771. If all looks good I'll publish the draft for #37 that stacks on this one.

I'll give some thought to addressing the report in #44. I'd like to keep the shipping cadence fast but we should be able to keep xsofy and let-go in sync and not regress due to drift. Let me know if you have any strong opinions or pre-existing plans there.

nooga (approved):

mparrett:
Thanks @nooga — appreciate the approval. While verifying this rebase I noticed the codebase no longer compiles against published let-go (v2.0.2), only against the fork. Filed nooga/let-go#132 with a couple of pragma shapes for surfacing version requirements declaratively; no action here, just flagging in case it's useful context.

PR #47, opened 2026-05-31
Faction-aware AI (friend/enemy partition) — #37

### LINKED ISSUE #37: Design proposal: faction-aware AI (friend/enemy concept)
Follow-up to #32. PR #35 adds `:friend` as a no-op stub in the rune vocabulary — the conjured creature still hunts the player, so the rune doesn't yet do anything observable. To make `:friend` (and any future allied-creature mechanic) real, the AI in `xsofy.ai` needs a faction concept it doesn't currently have. Filing this issue so the design gets discussed before anyone starts coding.

## Minimum needed for "marker only" to "rune actually works"

- **Target acquisition**: in AI, hostiles target `:friendly` creatures and the player; friendlies target hostiles.
- **Player melee**: bumping a `:friendly` creature shouldn't attack; probably block, or maybe swap places?
- **Player ranged / spells**: decide if AoE spells hit friendlies. Today the `r-damage` / `r-fire` / etc. cursors target every entity in a cell.
- **Render**: glyph tint or marker to distinguish player summons apart from a same-species hostile at a glance.

## Design Questions

A handful of these lean more balance/UX than implementation details, and it'd be good to get opinions before any PR surfaces:

1. **Faction model — boolean or richer?** `:friendly true` is enough for "player vs everyone else"; a `:faction` keyword would enable neutrals, rival factions, etc. later without revisiting. Slight up-front cost for future flexibility. I lean slightly toward boolean unless there's a clear idea of additional faction types in the near future.

2. **Lifespan.** Are conjured friendlies permanent until killed, or do they decay after N turns? Permanent is simpler; decay reads as "summons are a tactical resource." Slight lean toward decay-as-default, but least complexity option is permanent (aside from normal damage/death).

3. **Cap.** Can the player summon arbitrarily many friendlies (one per `:conjure` cast)? An untyped cap (e.g. "≤3 alive at once") prevents trivial steamrolling. I think we likely want a cap, but playtesting likely required to know what that cap is.

4. **Loot & XP.** If a friendly kills a hostile, who gets the drops? Options: 1. nothing drops 2. drops land at the kill site (player can pick up) or 3. friendly carries it (and you can presumably "ask" for it via an interaction). Carrying is enticing, and brings fond memories of Torchlight, but we should likely defer for an initial cut; drop at the kill site is intuitive and recommended.

5. **Cross-floor persistence.** When the player descends, do friendlies follow? If not, what happens to them — despawn, stay behind, save with the floor? Genuinely a tough call —  following is likely best but most complex.

6. **Friendly fire from the player.** If the player's `arc` or `area fire` spell sweeps a cell with a friendly, do they take damage? Most roguelikes lean "yes, you have to aim carefully" but it's worth a deliberate call.

7. **Regen.** Do friendlies use the same `regen-rate` curve as monsters (very slow), the player's curve, or none?

8. **Other ways to get friendlies?** Once the faction concept exists, items / runes / interactions could all produce one. Worth scoping whether this lands as "only `:friend` for now" or "the general mechanism."

## Proposed minimum-viable starting point

- `:friendly true` boolean on the entity (defer richer factions until needed)
- Target acquisition + bump-attack exclusion + glyph tint
- Friendlies are permanent until killed — later refinement through playtest
- No cap — later refinement through playtest
- AoE spells + ranged attacks hit friendlies (consistent with current cursor-set behavior; punishes lazy targeting)
- Friendlies don't follow up/down stairs — they save with their floor like other creatures (and don't act while the player is away, matching current floor-save semantics)
- No loot from friendly kills — keeps the player incentivized to do their own damage (downside: maybe less fun)

I'm happy to implement against whatever shape best fits the vision. Anything here you feel strongly about / want to do differently?

mparrett:
Closes #37 (minimum-viable cut).

PR #35 set `:friendly` on freshly conjured creatures, but the flag was inert —
friendlies still targeted the player, took bumps and AoE, and didn't follow
the descent. This PR implements the MVP shape sketched in #37's thread:

- **AI partitions targets by faction.** Hostiles hunt the player and any
  friendly; friendlies hunt hostiles. `:friend` (a stub in #35) now produces
  an entity that actually fights for you. `:conjure` alone still spawns a
  hostile, so `:friend` remains the rune that decides usefulness.
- **Player movement.** Bumping a friendly is a one-tick symmetric swap, not
  an attack. Single-target ranged (fire-bow + reach/skip filter) skips allies.
  AoE deliberately still hits them.
- **Render.** Cyan-blue glyph tint + `(ally)` sidebar marker; status tints
  still override so a burning ally reads as burning.
- **Follow on descent**, modeled as proposed: each friendly records its
  chebyshev distance from the stairs at descent time, then appears on the
  new floor after a delay proportional to that distance. The queue lives on
  the player so it bounces back if they ascend.

Two summon-side bugs surfaced and got fixed in passing: slime split clones
inherit `:friendly` (no more hostile shards from a friendly slime), and
`r-conjure` inherits the caster's `:ticks` (otherwise the summon's `:ticks 0`
dragged every low-tick actor forward, making hostiles "jump" tiles per
player step).

**Out of scope (per the thread):** weighted conjure / OP-summoning tuning
(happy to take it as a follow-up once surrounding pressure systems give us
something to tune against), leaders + orbit-following, loot routing, summon
caps, decay timers.

## Commits

8 commits, sequenced for review: AI partition → world/movement → render →
tests → two summon-side fixes (slime split, conjure ticks) → stairs-follow →
chore. Each is self-contained and passes the suite.

## Design notes

- **Faction flag is `:friendly` boolean** per "simple flag is enough" in #37. Open to renaming if you'd prefer `:faction :player`.
- **AoE deliberately does not filter friendlies.** `r-damage` / `r-fire` / `r-ice` / `r-arc` / `r-push` / `r-web` all hit allies — friendly fire is a feature, forces deliberate aim.
- **Out of scope** (per #37): leaders/orbit-following, loot routing, summon-count caps.

## Testing

- [x] 233 tests, 2145 assertions, 0 fail (baseline 226/2133).
- [x] Local CLI smoke test - conjured friendlies behave as designed

## Caveats
- Tested against let-go release v1.9.0
- Draft for your review pace; happy to split or reshape commits.

### DIFF
--- a/xsofy/ai.lg
+++ b/xsofy/ai.lg
@@ -150,23 +150,18 @@
 (def investigate-turns 6) ;; turns to search last known position
 
 (defn in-fov?
-  "Is this position in the player's current FOV? Friendlies share the
-   player's awareness — see docs/ai-design.md for the design call.
-   Used as the visibility proxy in find-target and the state machine."
+  "Pos in the player's current FOV. Friendlies share the player's
+   awareness — see docs/ai-design.md."
   [world pos]
   (contains? (or (:fov world) #{}) pos))
 
 (defn find-target
-  "Return the entity this actor should target, or nil. Friendlies seek
-   the nearest visible hostile; hostiles seek the player, or the
-   nearest visible friendly if one is closer. Actor itself must be in
-   FOV — mutual-visibility gate, same as the old can-see-player?
-   semantics. Returns nil when the actor has no eligible quarry, which
-   the state machine treats as 'idle' (fall through to sleep/wander).
+  "Entity this actor should target, or nil. Friendlies seek nearest
+   visible hostile; hostiles seek the player, or the nearest visible
+   friendly if closer. Actor must be in FOV.
 
    Ties on chebyshev distance break on (str :id) so equidistant
-   candidates are picked deterministically from (seed, action-log).
-   Same discipline as the rest of the engine."
+   candidates are picked deterministically from (seed, action-log)."
   [world entity]
   (let [pos (:pos entity)]
     (when (and pos (in-fov? world pos))
@@ -204,9 +199,8 @@
 ;; :investigate — moving to last known target position, then searching
 ;; :flee      — running away from target (low HP / burning)
 ;;
-;; "Target" is the player for hostiles, the nearest hostile for
-;; friendlies. find-target above does the lookup; everything else here
-;; treats target uniformly. Pre-faction code hardcoded :player.
+;; "Target" = find-target's return value. Player for hostiles, nearest
+;; hostile for friendlies.
 
 (defn update-ai-state [world entity-id]
   (let [entity (get-in world [:entities entity-id])
@@ -384,9 +378,9 @@
             target (find-target world entity)
             target-id (:id target)
             target-pos (:pos target)
-            ;; :investigate uses the LAST KNOWN position from :ai :target,
-            ;; not the live find-target lookup (the whole point is searching
-            ;; somewhere we no longer see them).
+            ;; :investigate uses LAST KNOWN from :ai :target, not the
+            ;; live find-target lookup (whole point is searching where
+            ;; we no longer see).
             last-known (get ai :target)]
         (if (nil? pos)
           world
@@ -398,9 +392,8 @@
 
             :hunt
             (if (nil? target)
-              ;; lost target this turn (e.g. friendly killed last hostile) —
-              ;; idle this tick, update-ai-state will rotate to investigate
-              ;; or sleep on the next pass.
+              ;; lost target (e.g. friendly killed last hostile); idle,
+              ;; update-ai-state rotates state next pass.
               world
               (case behavior
                 :ranged (hunt-ranged world entity-id pos target-id target-pos
--- a/xsofy/ai.lg
+++ b/xsofy/ai.lg
@@ -149,39 +149,82 @@
 (def wander-chance 30)    ;; % chance to wander each turn when sleeping
 (def investigate-turns 6) ;; turns to search last known position
 
-(defn can-see-player? [world entity]
-  (let [fov (:fov world)
-        pos (:pos entity)]
-    ;; simplified: if entity is in player's FOV, assume mutual visibility
-    (contains? fov pos)))
+(defn in-fov?
+  "Is this position in the player's current FOV? Friendlies share the
+   player's awareness — see docs/ai-design.md for the design call.
+   Used as the visibility proxy in find-target and the state machine."
+  [world pos]
+  (contains? (or (:fov world) #{}) pos))
+
+(defn find-target
+  "Return the entity this actor should target, or nil. Friendlies seek
+   the nearest visible hostile; hostiles seek the player, or the
+   nearest visible friendly if one is closer. Actor itself must be in
+   FOV — mutual-visibility gate, same as the old can-see-player?
+   semantics. Returns nil when the actor has no eligible quarry, which
+   the state machine treats as 'idle' (fall through to sleep/wander).
+
+   Ties on chebyshev distance break on (str :id) so equidistant
+   candidates are picked deterministically from (seed, action-log).
+   Same discipline as the rest of the engine."
+  [world entity]
+  (let [pos (:pos entity)]
+    (when (and pos (in-fov? world pos))
+      (let [es (vals (:entities world))
+            dist-to (fn [e] (distance pos (:pos e)))]
+        (if (ent/friendly? entity)
+          (->> es
+               (filter (fn [e]
+                         (and (ent/hostile-to-player? e)
+                              (:pos e)
+                              (in-fov? world (:pos e)))))
+               (sort-by (juxt dist-to (comp str :id)))
+               first)
+          (let [player (get-in world [:entities :player])
+                friendly (->> es
+                              (filter (fn [e]
+                                        (and (ent/friendly? e)
+                                             (:pos e)
+                                             (in-fov? world (:pos e)))))
+                              (sort-by (juxt dist-to (comp str :id)))
+                              first)]
+            (cond
+              (and player friendly (:pos player)
+                   (< (dist-to friendly) (dist-to player)))
+              friendly
+              (and player (:pos player)) player
+              :else friendly)))))))
 
 ;; --- AI State Machine ---
 ;; States: :sleep, :wander, :hunt, :investigate, :flee
 ;;
-;; :sleep     — standing still, might wander or wake on seeing player
-;; :wander    — roaming randomly, alert to player presence
-;; :hunt      — chasing the player, attacks when adjacent
-;; :investigate — moving to last known player position, then searching
-;; :flee      — running away from player (low HP)
+;; :sleep     — standing still, might wander or wake on spotting target
+;; :wander    — roaming randomly, alert to target presence
+;; :hunt      — chasing target, attacks when adjacent
+;; :investigate — moving to last known target position, then searching
+;; :flee      — running away from target (low HP / burning)
+;;
+;; "Target" is the player for hostiles, the nearest hostile for
+;; friendlies. find-target above does the lookup; everything else here
+;; treats target uniformly. Pre-faction code hardcoded :player.
 
 (defn update-ai-state [world entity-id]
   (let [entity (get-in world [:entities entity-id])
-        player (get-in world [:entities :player])]
-    (if (or (nil? entity) (nil? player) (nil? (:pos entity)) (nil? (:pos player)))
+        target (when entity (find-target world entity))]
+    (if (or (nil? entity) (nil? (:pos entity)))
       world
       (let [ai (:ai entity)
             state (get ai :state :sleep)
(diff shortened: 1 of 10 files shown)


nnunley (commented):
Nice PR — clean faction partition and the bump-swap tick accounting is well-reasoned. Inline comments flag three things that need fixing before merge; all are small. (Two are determinism regressions relative to the repo's seed-reproducibility work; one is a latent divide-by-zero.)

Separately, not line-level: the core invariant *"a hostile with no nearer friendly targets the player"* isn't pinned by any test (here or in base — the existing AI tests are only no-throw/invariant props), so a regression in the `find-target` rewrite could pass the whole suite. Worth one deftest.

xsofy/ai.lg
```diff
@@ -149,39 +149,73 @@
...
+    (when (and pos (in-fov? world pos))
+      (let [es (vals (:entities world))
+            dist-to (fn [e] (distance pos (:pos e)))]
+        (if (ent/friendly? entity)
+          (->> es
+               (filter (fn [e]
+                         (and (ent/hostile-to-player? e)
+                              (:pos e)
+                              (in-fov? world (:pos e)))))
+               (sort-by dist-to)
```
nnunley: Non-deterministic target selection. This `(->> (vals (:entities world)) (filter …) (sort-by dist-to) first)` (and the matching one at L181) picks among **equidistant** candidates by map-vals iteration order, with no stable tiebreak — so creature targeting isn't reproducible from `(seed, action-log)`. Same class as the existing `:wait`/creature non-determinism. Add a secondary key:

```clojure
(sort-by (juxt dist-to :id) …)
```

xsofy/world.lg
```diff
@@ -496,8 +497,97 @@
...
+        (let [ring (vec (for [dx (range (- r) (inc r))
+                              dy (range (- r) (inc r))
+                              :when (or (= (math/abs dx) r)
+                                        (= (math/abs dy) r))
+                              :let [nx (+ ox dx) ny (+ oy dy)]
+                              :when (terrain/walkable? terrain width height nx ny)
+                              :when (nil? (ent/creature-at world [nx ny]))]
+                          [nx ny]))]
+          (if (seq ring)
+            (nth ring (rand-int (count ring)))
```
nnunley: Host RNG breaks seed-reproducibility. `(rand-int (count ring))` is non-deterministic, so any turn that materializes an ally can't be replayed from a seed. This PR adds no `det/` calls. Route it through the deterministic helpers salted on `:seed`, e.g.:

```clojure
(det/pick world [:conjure :ally-spawn r] ring)
```
(or `det/int-in` for the index), threading the world so `:seed` advances.

xsofy/world.lg
```diff
@@ -1422,6 +1515,31 @@
...
+      ;; :ticks stay put and run-until-player-turn gives them a free
+      ;; turn (net: 2-tile ally move per player step). /speed preserves
+      ;; speed-2 allies' advantage.
+      (let [c (ent/creature-at world [nx ny])]
+        (and c (:friendly c)))
+      (let [bump (ent/creature-at world [nx ny])
+            bname (or (:item-name bump)
+                      (name (or (:template bump) (:id bump))))
+            bump-speed (or (get-in bump [:ai :speed]) 1)
+            bump-cost (quot move-cost bump-speed)]
```
nnunley, line 1542: Latent divide-by-zero. If an ally is templated `:ai {:speed 0}`, `(quot move-cost 0)` throws. `bump-speed` defaults to 1 when absent, but an explicit `0` slips through. Guard with `(max 1 bump-speed)`:

```clojure
bump-cost (quot move-cost (max 1 bump-speed))
```

xsofy/test/combat_prop.lg
```diff
@@ -173,3 +173,95 @@
...
+    (if (empty? creatures)
+      true
+      (g/world-invariants?
+        (ai/ai-act world (first creatures)
+                   w/melee-attack w/ranged-attack)))))
+
+(c/defprop process-creatures-with-friendlies-preserves-invariants
+  [world g/gen-rich-world-with-friendly]
+  (g/world-invariants?
+    (ai/process-creatures world w/melee-attack w/ranged-attack)))
```
nnunley, line 267: Two tests for the gaps noted above — both self-contained (reuse the `minimal-with-player` helper added in this PR), drop in right here after the faction props.

**1. Core invariant — hostile targets the player** (pinned nowhere today):

```clojure
(deftest hostile-targets-player-when-no-friendly
  (testing "find-target for a hostile with no visible friendly returns the
            player. The core invariant the faction rewrite must preserve."
    (let [world (-> (minimal-with-player [3 3])
                    (assoc-in [:entities :goblin-1]
                      {:id :goblin-1 :template :goblin :pos [5 3]
                       :ticks 0 :status {} :statuses {}
                       :ai {:state :hunt :behavior :hunter :speed 1}
                       :body {:hp 15 :max-hp 15 :strength 2 :armor 0}})
                    w/update-fov)
          target (ai/find-target world (get-in world [:entities :goblin-1]))]
      (is (some? target) "hostile has a target")
      (is (= :player (:id target)) "and it is the player"))))
```

**2. Deterministic tiebreak** — pairs with the `(juxt dist-to (comp str :id))` fix flagged on the `sort-by` lines above. Verified it fails on the current sort (picks the wrong one on a tie) and passes with the fix. `:fov` is set explicitly so only the tiebreak is under test, not FOV geometry:

```clojure
(deftest find-target-tiebreak-is-deterministic
  (testing "Among equidistant visible candidates, find-target picks the same
            one regardless of entity-map insertion order (stable :id tiebreak)."
    (let [ally  {:id :ally-1 :template :rat :pos [4 4] :friendly true
                 :ticks 0 :status {} :statuses {}
                 :ai {:state :hunt :behavior :hunter :speed 1}
                 :body {:hp 6 :max-hp 6}}
          gob-a {:id :goblin-a :template :goblin :pos [4 6]   ; distance 2 from ally
                 :ticks 0 :status {} :statuses {}
                 :ai {:state :hunt :behavior :hunter :speed 1}
                 :body {:hp 15 :max-hp 15}}
          gob-b {:id :goblin-b :template :goblin :pos [6 4]   ; distance 2 from ally
                 :ticks 0 :status {} :statuses {}
                 :ai {:state :hunt :behavior :hunter :speed 1}
                 :body {:hp 15 :max-hp 15}}
          base (-> (g/minimal-stone-room)
                   (assoc :fov #{[4 4] [4 6] [6 4]})
                   (assoc-in [:entities :player]
                     {:id :player :template :player :pos [9 9]
                      :ticks 0 :status {} :statuses {} :inventory []
                      :body {:hp 30 :max-hp 30}}))
          w-ab (-> base (assoc-in [:entities :goblin-a] gob-a)
                        (assoc-in [:entities :goblin-b] gob-b)
                        (assoc-in [:entities :ally-1] ally))
          w-ba (-> base (assoc-in [:entities :ally-1] ally)
                        (assoc-in [:entities :goblin-b] gob-b)
                        (assoc-in [:entities :goblin-a] gob-a))
          t-ab (ai/find-target w-ab (get-in w-ab [:entities :ally-1]))
          t-ba (ai/find-target w-ba (get-in w-ba [:entities :ally-1]))]
      (is (some? t-ab) "a target is found")
      (is (= (:id t-ab) (:id t-ba)) "target is independent of insertion order")
      (is (= :goblin-a (:id t-ab)) "deterministic pick: lowest :id wins on a tie"))))
```

mparrett:
@nnunley Thanks for reviewing! Fixes for the current set of findings coming shortly. Feel free to add more if you spot anything.

xsofy/ai.lg
mparrett: Fixed in `aaaf40a` (amended into the source commit). Used `(juxt dist-to (comp str :id))` per your follow-up — the `str` coercion keeps ordering stable regardless of `compare` semantics for keywords. Pinned by the new `find-target-tiebreak-is-deterministic` deftest.

xsofy/world.lg
mparrett: Fixed in `9d5c397` (amended into the stairs-follow commit). `find-spawn-near` now returns `[pos world']` and routes the ring pick through `(det/pick (det/with-context world [:ally-spawn r]) ring)` — salted on `r` so each fallback radius gets its own sub-chain. `materialize-pending-ally` destructures the return and threads the advanced `:seed` forward.

xsofy/world.lg
mparrett: Fixed in `bbbbcaf` (amended into the bump-swap commit). `bump-speed (max 1 (or (get-in bump [:ai :speed]) 1))` — the `max 1` catches the explicit-zero case the `or` couldn't.

xsofy/test/combat_prop.lg
mparrett: Both added verbatim in `bdce029` (amended into the test commit). Suite: 235 tests / 2150 assertions / 0 fail. Good call on the placement — `combat_prop.lg`'s existing `minimal-with-player` was the right home; thanks for writing them out.

nnunley (approved):

Issue #4, opened 2026-05-14
Cannot use enchantment scroll on item in slot j


Oakwhisper:
When trying to use an enchantment potion on the item in slot j, it instead moves the cursor around. There should be some option to use enter to select the currently highlighted item.

PR #48, opened 2026-05-31
PR6: Commit DAG (content-addressed worlds, branching, time-travel, diegetic GC)

nnunley:
## Summary

Implements **PR6** of the deterministic-world design: a content-addressed commit DAG layered on top of the existing world model. Every dispatched turn can produce a commit; HEAD is tracked; branching and time-travel work; a deterministic "the world forgets" GC bounds memory. Demonstrated end-to-end by a dev-only `(debug/undo)`.

Built on the *shipped* reality (the world carries its own `:seed`/`:action-log`/`:turn`; PR5 deferred the wrapper) rather than the spec's assumed game-state wrapper.

### What's new
- **`xsofy/serialize.lg`** — canonical, type-tagged byte encoding of a world: stable under map key order, injective on shape, and crucially handles the types real worlds contain (`:fov` Set, `:terrain` byte-array). Foundation for content-addressing.
- **`xsofy/dag.lg`** — `commit-id` (= xxh3-64 of canonical bytes); game-state wrapper `{:world :dag :parents :head :refs :budget}`; `commit`, `peek`, `switch-head`, `walk-parents`, `branch!`; deterministic `gc` (oldest-by-turn, commit-id tiebreak; pins HEAD + named refs only, so distant ancestors are forgettable).
- **`xsofy/dispatch.lg`** — `dispatch-gs` (one commit per state-mutating turn + GC; UI actions no-op) and `replay-gs`.
- **`xsofy/debug.lg`** — dev-only `undo` (walks HEAD back one commit).

### Scoped out (intentional, per spec)
- No `policy.lg` / `:branches` registry (non-`:isolated` policies are additive future work).
- The live game loop (`main.lg`) is **not** migrated to game-state yet — PR6 ships the DAG exercised via `dispatch-gs`/`replay-gs`/`undo`, as the spec describes.
- **PR7** (save/load + `rewind`/`peek`/`delay` spells) is a follow-on.

## Test Plan
- [x] `lg xsofy/test/run.lg` — 248 tests, 0 fail (incl. ~22 new: serializer stability/injectivity/sets/arrays, commit-id full-world coverage, DAG ops, deterministic GC, replay==live, undo)
- [x] `lg -b /tmp/x main.lg` — game still compiles (exit 0)
- [x] Each task reviewed (spec-compliance + code-quality) + final whole-PR review

## Notes for reviewers
- World value-equality via `=` is unreliable in let-go (worlds contain a byte-array → `(= w w)` is false); world comparisons go through `commit-id`. 
- Found (out of scope, logged): `:wait` dispatch is non-deterministic in shipped code (creature path; PR4 migration incomplete) — determinism tests use pure actions; worth a PR4-completion follow-up.
- `dag/peek` intentionally shadows `clojure.core/peek` (harmless namespaced warning), matching the spec's `dag/peek` naming.


mparrett:
@nnunley Ran this by Opus 4.7 and did some cross-checking. Read without the referenced spec doc (`docs/det-*.md`) and relying on the PR body for scope). Admittedly, beyond my ability to discern at a high level of detail, but the overall shape of the feedback seems reasonable. Here's what surfaced:

### `dispatch-gs` silently drops UI-action state

`world/update-world` is not a no-op for the UI keys in `ui-actions` — it sets `:screen` (and sometimes `:menu-action`) on the world:

```clojure
;; xsofy/world.lg
:messages    (assoc world :screen :messages)
:help        (assoc world :screen :help)
:inventory   (assoc world :screen :inventory)
:equip-menu  (assoc world :screen :quick-menu :menu-action :equip)
…
```

But `dispatch-gs` computes `w'` and discards it on the UI branch:

```clojure
(defn dispatch-gs [gs action]
  (let [w' (dispatch (dag/head-world gs) action)]
    (if (replay-action? action)
      (dag/gc (dag/commit gs w'))
      gs)))                       ; ← w' computed, then thrown away
```

`dispatch-gs-ui-action-does-not-commit` only asserts `(= head₀ head₁)` and `(= 1 (count :dag))`, so the test passes on the broken behavior — it never checks that the screen actually toggled. The "no commit" property is right, but the world update needs to propagate, e.g. `(assoc gs :world w')` on the UI branch. Latent for now since `main.lg` isn't on `dispatch-gs` yet, but it'll bite on the migration.

### GC eviction-policy docstring/keyword vs. implementation

`:budget` advertises `:evict-policy :oldest-by-distance` and the `gc` docstring says "largest distance from HEAD," but the sort key is `[(commit-turn gs id) (str id)]` — i.e. lowest `:turn` first, which is what the PR body honestly describes. For linear play these coincide; for branches off an older commit they don't (a turn-100 world branched off turn-2 ranks "young" by turn but is graph-distant). Either rename the keyword to `:oldest-by-turn` and align the docstring, or actually compute distance.

### `walk-parents` past a GC'd parent

Docstring: "nil if the chain runs out (root reached or a link was GC'd)." Actual behavior: when HEAD's parent has been evicted, `(walk-parents gs 1)` returns the evicted id (still stored as a *value* on HEAD's `:parents` entry, even though the evicted id's own `:parents` row was dissoc'd). `switch-head` and `undo` defensively `peek`-check, so no functional bug today — just contract drift worth aligning.

### Perf: worth benching before `main.lg` migrates

`canonical-bytes` walks every world, every committing turn: `:terrain` (byte-array, w*h bytes), `:lighting` (int-array, ~w*h * 9 bytes once each int is TAG-INT+i64-be), the full `:fov` set, and `:entities`. For 80×50 that's ~36KB just for lighting, built via `into`/`conj` on a persistent vector, then `xxh3-64` on the long (>240) pure-lisp path. xxh3 is cheap in C but the pure-lisp long path is the most expensive routing in `xsofy.hash`. Not asking to redesign — but it'd be good to see a turn-time before/after on a representative floor before PR-N puts this on the hot path. A sub-tree-hash approach (memoize per array identity, combine child hashes) would skip the unchanged-terrain/lighting work without losing the content-address property.

### Smaller

- `serialize`'s set/map branches sort by `(vec (emit elem))` and then `emit-into`s again — `emit` runs twice per element. For `:fov` (hundreds of cells) this is real. Compute the keys once via `(map (juxt emit identity) …) → sort-by first → unpack`.
- `branch!` is pure (returns `gs'`); the bang is misleading vs. let-go/Clojure convention.

nnunley:
Thanks for the review. The hash calculation was a good catch. Given our deterministic world, we really don't need to do content hashing. We just need to capture the seed after the current player action has been taken. Everything else is deterministic based on player(s) input. Deeper responses and code coming shortly.

nnunley:
As an aside, the xx3 hash was going to be shifting to native hashing once 1.9.0 is officially released, so the calculation should go from ms to ns. The local version was a stopgap to allow me to use a strong deterministic hash to provide the appearance of randomness.

nnunley:
All six are addressed in the latest push (force-updated to a single commit `PR6: commit DAG — action-log storage + replay-on-demand`). The perf point (#4) drove a design rework, so a couple of these are resolved structurally rather than patched:

**1. `dispatch-gs` dropped UI-action state** — fixed. UI actions now propagate the world update without committing: in the action-log model, a replay action logs a node + advances HEAD; a UI action does `(assoc gs :world w')` only (no node). So the screen toggle takes effect and is correctly *not* logged (replay only replays state-mutating actions). Test now asserts `(= :help (:screen (head-world gs1)))`.

**2. GC policy keyword/docstring vs. impl** — fixed. Renamed `:oldest-by-distance` → `:oldest-by-turn` and aligned the docstring (it sorts by `:turn`, which is turn-difference, not graph distance). GC itself changed in the rework — it now advances the replay *origin* and forgets the pre-origin past.

**3. `walk-parents` past a GC'd parent** — fixed. It now returns `nil` when the nth ancestor is no longer materialized, matching the docstring; callers can treat a non-nil result as still-reachable.

**4. Per-turn `canonical-bytes` (~65ms)** — resolved by design, with data. Benchmarked: `commit-id` was ~65ms/turn (39ms serialize + 25ms xxh3 on 52KB) — far over a frame budget, untenable on the hot path. Since the world is fully determined by input, the rework **stops storing/hashing worlds per turn**:
  - `commit-id` is now the rolling `:seed` hash (O(1), already advanced each turn).
  - commits store only the **action log** (`:nodes` = parent/action/turn); past worlds **regenerate by replay** on demand (~0.35s/level, measured).
  - the full canonical hash survives as `checkpoint-hash` — **optional**, only at non-history merge points (save/load reconstruction, external injection), never per turn. The sub-tree-hash idea is noted there for if/when it's needed.

**5. `serialize` double-emit on sets/maps** — fixed. Each element is encoded once (sort by the precomputed bytes, reuse them) instead of emitting twice. (serialize.lg's only consumer now is `checkpoint-hash`.)

**6. `branch!` bang** — fixed. Renamed to `branch` (it's pure, returns a new game-state); it also now takes a real action rather than an arbitrary fn, keeping it replay-consistent.

Also relocated `undo` from `debug.lg` into `dag.lg` (it's a DAG op, not crash/logging plumbing) — which additionally broke the `dag ↔ dispatch` require cycle the rework introduced.

255 tests pass; native compile clean. Happy to bench `checkpoint-hash` on a representative floor if useful before any save/load wiring.

nnunley:
Refinement on the GC (supersedes the "advances the origin / forgets" note above): `gc` is now **reachability pruning** — it keeps HEAD's chain and every named ref's chain back to the origin, and drops the rest (abandoned rewind-and-diverged branches). Refs pin their timelines. It's safe by construction: a pruned node has no surviving descendant, so no reachable commit ever loses its replay chain and common parents are always kept. It's also explicit, not per-turn (forward play creates no unreachable nodes, and re-walking the chain each turn would be O(turns²)).

Deliberately deferred: forgetting *reachable* history (origin advancement, or a "burn the past" mechanic). In the replay-from-origin model, cutting a reachable ancestor severs the replay chain of all its descendants — so any such removal must first **anchor** the surviving branches (materialize + store a world checkpoint at the cut), else a common parent becomes unreachable-without-an-anchor. That's a policy for later; the action log is tiny, so reachability pruning alone suffices now.

nooga (approved):

PR #51, opened 2026-05-31
fix: keep title & death-screen particles animating (and calm the pulse)

nnunley:
## Summary
The title and death screens froze their rune particles until a key was pressed — `term/read-key` is blocking, so frames only advanced around input. Now keys are read on a one-shot goroutine and the loop selects between that channel and a frame timer, so the particles animate continuously while waiting.

- **title** — `read-key-chan` / `make-key-poller` / `drive-until-key` helpers; animate until a key instead of holding a static final frame.
- **death** — thread a frame counter into `render-death-screen`; pulse the runes via `rune-brightness` while waiting for input.
- **pulse cadence** — 120ms/frame (was 80ms) for a calmer particle motion; input latency at these screens stays imperceptible.

## Test plan
- [x] `lg xsofy/test/run.lg` — 237 tests pass (incl. new `drive-until-key` control-loop coverage in `title_test`)
- [x] `lg -b /tmp/x main.lg` — compiles (exit 0)
- [x] Verified in a PTY that both screens keep animating with no keypress, past the old freeze point

### DIFF
--- a/main.lg
+++ b/main.lg
@@ -428,12 +428,18 @@
                      (recur w w (term/size))))))
 
          :death
-         (let [result (loop [scroll 0]
-                        (render/render-death-screen world scroll)
-                        (let [k (term/read-key)]
+         (let [poll! (title/make-key-poller 120)  ;; calmer death-screen pulse (was 80ms)
+               result (loop [scroll 0 frame 0]
+                        ;; animate the death screen continuously; the runes keep
+                        ;; pulsing while we wait for a key instead of freezing.
+                        (let [[k nf] (title/drive-until-key
+                                       frame
+                                       (fn [f] (render/render-death-screen world scroll f))
+                                       poll!)]
                           (cond
                             (or (= k "n") (= k "\r") (= k "\n") (= k " "))
                             :new-game
+                            (= k :eof) (assoc world :running false)
                             (= k "\u001b")
                             (if *in-wasm*
                               (do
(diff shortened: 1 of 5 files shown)


nooga (approved):

mparrett:
Worth flagging: let-go has `term/key-pending?` (non-blocking peek primitive, landed in #118), which I used in xsofy #16 for the same "advance state while waiting for input" shape. The polled-peek version would be roughly:

```clojure
(loop [frame 0]
  (render frame)
  (if (term/key-pending?)
    (handle (term/read-key))
    (do (Thread/sleep frame-ms)
        (recur (inc frame)))))
```

Tradeoff: your channel/select shape is more composable if `make-key-poller` / `drive-until-key` are stepping stones toward multiplexing other sources (network, multi-timer). Polled-peek skips the goroutine lifecycle and channel-close test surface — simpler if this is the one site, and would let `title.lg` shed most of the new helpers.

Not a request to change anything — just surfacing the option since the primitive's already in `term/`.

mparrett:
@nnunley FYI in case it's useful for adjacent work: I just filed nooga/let-go#165 for the main game loop's SIGWINCH handling, and the PR body explains why I went with a self-pipe wake-byte rather than extending the channel-poll pattern from this PR. Short version — WASM's `Atomics.wait` blocks the worker thread synchronously, so a `(go (term/read-key))` parked in `read-key` prevents the parent's `alts!` from ever being scheduled; the channel decoupling is illusory on that target.

I haven't actually run #51 under WASM (your test plan calls out PTY only), so this is reasoned from the constraint rather than measured. If you've tested in browser and it works, I'd love to hear about it. There may be Go-WASM scheduler subtleties I don't know about.

The #51 pattern reads as exactly right for native title/death animation regardless. Those contexts aren't on the hot path of "WASM worker also doing other work," so the trap I'm worried about doesn't apply to where you actually use the helpers.

Issue #58, opened 2026-06-04
ci: deploy Pages from version tags instead of main


mparrett:
### Recommendation

Switch Pages deploys from `main` HEAD to tags matching `v*`, and cut `v0.0.1` at `78e2350`. One-line workflow change; reversible; restores a working live demo today without blocking any in-flight work on `main`.

### Why

Every push to `main` deploys to `nooga.github.io/xsofy/`. A regression that doesn't fail CI (e.g. runtime or rendering) reaches users immediately, and fixing the live demo today means either reverting on `main` or shipping a forward fix — both with no buffer for the public URL while that lands.

The currently-deployed bundle (`ffebbb3`) shows visible issues on first load: frozen title, map redraw glitches. The broader class of pain — "I hit a broken state, which version was that?" — is also the underlying ask behind #44.

Under this PR, `main` stays the dev tip. Tags become the public release channel: the demo only moves when you decide it should.

### Changes

```diff
 on:
-  push:
-    branches: [main]
+  push:
+    tags: ['v*']
   workflow_dispatch:
```

`workflow_dispatch` is retained for manual deploys from the Actions tab. Everything else in the workflow is unchanged. Until a `v*` tag exists, Pages keeps serving the last deploy (`ffebbb3`); tagging triggers a fresh build.

### First tag

`78e2350` ("Cancel auto-action on queued keystroke (#16)", 2026-05-29) is the most recent commit I've manually verified plays cleanly against both `let-go` `v1.9.0` (current `@latest`) and `let-go` `main`. Six commits sit between it and current `main` — the regression is somewhere in there, but bisecting is independent work and shouldn't gate restoring the demo.

```
git tag -a v0.0.1 78e2350 -m "v0.0.1 — first tagged release"
git push origin v0.0.1
```

`v0.0.x` reads as "first deliberate snapshot" without implying API stability; bumping to `0.1.0` / `1.0.0` is your call when a milestone warrants it.

### Follow-ups (separate work)

- Bisect `78e2350..ffebbb3` and revert/fix. Active design work (#55, #56) sits in this range — it's possible the regression rides along with something not yet finished landing, in which case waiting is also viable.
- Optional: visual/runtime smoke gate in CI strong enough to catch the kind of regression that escaped here.
- Optional, later: auto-tag-on-green once that gate exists — flips back to "trunk = release" with a real quality bar.

### Alternatives considered

[...]

PR #60, opened 2026-06-04
ci: add browser smoke gate (Part C-lite) before Pages deploy

mparrett:
Browser smoke gate (Part C-lite from the discussion in #58). Stacks on #59's tag-based deploy trigger.

Adds a deterministic boot/title-render gate before the `upload-pages-artifact` step. The smoke launches headless Chromium against the built `dist/`, waits for the literal "Press any key" subtitle from `xsofy.title` to appear in the rendered terminal text within 10s, and fails the deploy if it doesn't (or if any `console.error` / uncaught exception fires).

## Why "Press any key" as sentinel

It's a literal string in `xsofy/title.lg`, so the assertion is deterministic across the random seeding that varies room layout, enemies, and quest names every boot. Golden-snapshot row comparison was tested locally and is brittle for this reason — three boots of the same dist produce three completely different worlds.

## Empirical validation

- `xsofy@78e2350` (the v0.0.1 baseline): 3/3 pass, sentinel visible at ~5.3s
- `xsofy@ffebbb3` (current tip): 3/3 fail — sentinel never renders within 10s. "Press any key broken on tip" is part of the visible regression class.

## Implementation notes

- Playwright. `tests/browser/smoke.mjs` + `package.json` + `package-lock.json` (committed for reproducibility via `npm ci`).
- Only `console.error` and uncaught exceptions are fatal — `console.warning` was intentionally dropped during local review since browser/CDN warnings are too noisy for a deploy gate.
- Failed runs upload `smoke-failure.png` as a workflow artifact.
- ~30s added to deploy CI on cold install.

Doesn't catch the full xterm-render-correctness class — that's Part B territory (separate work, deferred).

Refs #58.

### DIFF
--- a/.github/workflows/deploy-pages.yml
+++ b/.github/workflows/deploy-pages.yml
@@ -45,7 +45,7 @@ jobs:
       - name: Install smoke dependencies
         working-directory: tests/browser
         run: |
-          npm install --no-audit --no-fund
+          npm ci --no-audit --no-fund
           npx playwright install --with-deps chromium
 
       - name: Browser smoke (must boot to title screen)
--- a/.github/workflows/deploy-pages.yml
+++ b/.github/workflows/deploy-pages.yml
@@ -37,6 +37,29 @@ jobs:
       - name: Patch COI startup guard
         run: let-go tools/patch_wasm_coi.lg
 
+      - name: Set up Node (smoke gate)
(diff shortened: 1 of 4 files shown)


nooga (approved):
Pretty crafty!

nnunley (approved):

Issue #56, opened 2026-06-02
Responsive UI event loop: composable, replay-safe input + cosmetic animation


nnunley:
## Problem
The determinism refactor made `game-loop` purely **input-driven** (blocking `read-key`) to protect bit-identical `(seed, action-log)` replay — which removed wall-clock and **froze idle animation** (title/death particles). In the browser it's worse: wasm `read-key` uses `Atomics.wait`, and Go-wasm is single-threaded, so a blocking read freezes the **whole** runtime, including any clock goroutine.

## Proposed design
A small, composable, Rx-flavored event loop (`xsofy/ui.lg`, extending the existing module):

- **Poll-based source contract** (never blocks while idle → browser-safe): `input-source` (`key-pending?` then `read-key` only when ready), a cosmetic `clock-source`, a generic `poll-source`, and a ctx-aware-only `chan-source` seam for future async sources.
- **`run-loop`** merges sources → optional transducer `xform` → `step` (a `scan`) → `render` subscriber; exit via `(reduced …)`. Paced through the existing `pause!` (wasm-yielding, ≥30ms floor).
- **Two orthogonal streams.** Loop state is `{:world :ui}`: deterministic input drives `dispatch` (seed + `:action-log`); **cosmetic frame ticks advance `:ui` only and never touch the seed/log** → replay stays bit-identical regardless of animation.
- **Modal & multi-step menus** (e.g. `apply → select item → optional target`) compose as **nested `run-loop`s** that return an assembled action; the whole flow is pure-UI, only the final action is dispatched/replayed.
- **Backend-agnostic core** — terminal (native + wasm) now; a 2D graphical backend (incl. Switch) reuses the same loop/`step` with different sources + render later.

## Spec
Full design + a 6-issue pushback review: **#55** (`docs/responsive-ui-event-loop-design.md`).

## Rollout
- **PR1** — `xsofy/ui.lg` event-loop surface + tests (incl. wasm pacing smoke test)
- **PR2** — title + death screens (supersedes #51)
- **PR3a** — core play-loop swap
- **PR3b** — modal menus + multi-step flows (nested run-loops)
- **PR4** — `docs/terminal.md` 'Responsive UIs' section

**Relates to:** supersedes #51 · depends on #52 (shared `screenfx`) · builds on the determinism work (#53/#54).

mparrett:
Status sweep as of 2026-06-19, checking current `origin/main` plus the merged PRs:

**Landed on `main`:**
[...]

PR #61, opened 2026-06-05
fix(title): WASM-safe input poll via key-pending? (deploy regression)

mparrett:
## Summary

- Bisected the deploy regression visible on https://nooga.github.io/xsofy/ to **3ce4813** ("fix: keep title and death-screen particles animating without input"). The new `make-key-poller` uses `(go (term/read-key))` + `async/alts!` against a timeout. In WASM, `term/read-key` calls `Atomics.wait` on the SharedArrayBuffer, which **blocks the entire worker thread** (not just the calling goroutine) — freezing the timer goroutine backing `async/timeout` and pinning the animation loop at frame 0 forever. The frame-30-gated \"Press any key\" subtitle never renders.
- Rewrites `make-key-poller` to use the let-go `term/key-pending?` primitive (provided precisely for this idiom — see term.go docstring) plus the existing `pause!`. Drops the now-unused `read-key-chan`. `drive-until-key` is unchanged; the `title_test` seam stays valid.
- Native behavior is functionally equivalent. The death screen (uses the same poller via `main.lg`) gets the same fix.

## Why the goroutine pattern broke specifically in WASM

`term/read-key` source:
```go
// pkg/rt/term_wasm.go
atomics.Call(\"wait\", keyInt32, 0, 0)   // ← blocks worker thread, not just goroutine
```

Native `read` is goroutine-cooperative; WASM `Atomics.wait` is not. The let-go term module ships `key-pending?` (non-blocking Atomics.load probe) for exactly this idiom — and the native `pkg/rt/term.go` comment even calls it out: *\"makes the `(when (key-pending?) (read-key))` idiom honest for animation / poll loops.\"*

## Validation

Built bundle of each at `78e2350` (last-known-good baseline = v0.0.1), `ffebbb3` (broken deploy tip), and this PR's tip; ran the Part C-lite Playwright smoke from #60 10 times each.

| | v001 (good) | ffebbb3 (broken) | this PR |
|---|---|---|---|
| Smoke pass (n=10) | 10/10 | 0/10 | **10/10** |
| Boot ms (mean) | 734 | timeout | 659 |
| Sentinel ms (mean) | 5,739 | never paints | 6,624 |
| Gameplay frames (5 moves) | 5–6/6 distinct | 0 — can't advance | **6/6 distinct** |
| Console errors / page exceptions | 0 / 0 | 0 / 0 | 0 / 0 |
| Title JS heap Δ (MB) | 1.80 | 1.35 | **0.62** |

Sentinel delta vs v001 is +885ms, which is **less than** the intentional +1,200ms cadence change in 3ce4813 (30 frames × (120ms − 80ms) of \"calmer particle motion\"); no per-frame overhead is added by the fix. JS heap allocation is strictly lower than both — no `(go ...)` channel allocs per keypress arming.

Headless key-spam against the broken build (focused `#terminal`, 50+ space presses over 25s) never advances the title screen, confirming the deadlock is total — not a timing race.

Full lg test suite (260 tests, 2,227 assertions) passes against the fix branch.

## Test plan

- [x] Local Playwright smoke (`tests/browser/smoke.mjs` from #60): 10/10 pass; sentinel renders at 6.3–7.3s
- [x] Local lg test suite: 260/260 pass, 0 fail
- [x] Manual Safari on phone (Tailscale-served): title → \"Press any key\" → space → world renders → hjkl moves track 1:1
- [x] Compared perf vs v001 baseline: matches within design parameters, lower heap
- [ ] CI test.yml passes on the PR (lg test suite cell)
- [ ] On merge → tag → deploy, the Part C-lite gate (#60) will independently re-validate the sentinel before the bundle goes live

## Notes

- Once this merges, a fresh tag on top of main (v0.0.2 or moved v0.0.1) picks up the new deploy workflow (#59) + smoke gate (#60) naturally — no cherry-pick or history rewrite needed to recover the live URL.
- The bisect range was 6 commits (`78e2350..ffebbb3`); `ffebbb3` itself is docs-only (#55) so the effective range was 5 code commits. `3ce4813` reproduced FAIL on first spot-check.

### DIFF
--- a/xsofy/screenfx.lg
+++ b/xsofy/screenfx.lg
@@ -15,37 +15,28 @@
 
 ;; --- Non-blocking input for animated screens ---
 ;; `term/read-key` is blocking, so a screen that calls it directly freezes its
-;; animation until a key is pressed. To keep particles running, we read keys on
-;; a goroutine and select between that channel and a frame timer.
+;; animation until a key is pressed. To keep particles running we sleep one
+;; frame, then peek the input via the non-blocking `term/key-pending?` and only
+;; consume a key when one is queued.
 ;;
-;; The reader is ONE-SHOT: it reads a single key, delivers it, and exits. A
-;; goroutine blocked inside `read-key` cannot be cancelled, so a looping reader
-;; would survive past the screen and steal the next keypress from the game loop.
-;; One-shot + lazy re-arming guarantees at most one reader is ever blocked on
-;; stdin, and none is left behind once a screen exits on a key.
-
-(defn read-key-chan
-  "Spawn a goroutine that reads exactly one key and delivers it on the returned
-   channel, then exits. EOF (nil from read-key) is delivered as :eof."
-  []
-  (let [ch (async/chan)]
-    (go (async/>! ch (or (term/read-key) :eof)))
-    ch))
+;; Why not a goroutine + async/alts! on read-key: in WASM, `term/read-key` calls
+;; `Atomics.wait` on the SharedArrayBuffer, which blocks the entire worker
+;; thread (not just the calling goroutine). That freezes the timer goroutine
+;; backing `async/timeout`, so `alts!` never returns nil and the animation loop
+;; never advances past frame 0 — exactly the "Press any key" never-renders
+;; symptom observed on deployed builds. `key-pending?` is the let-go primitive
+;; provided for this idiom.
 
 (defn make-key-poller
-  "Return a 0-arg poll fn for animation loops. Each call waits up to `ms` for a
-   keypress: returns the key string, :eof, or nil if none arrived in time.
-   Arms a one-shot reader lazily and disarms once a key is delivered, so when a
-   screen stops polling (after its terminating key) no reader is left blocked."
+  "Return a 0-arg poll fn for animation loops. Each call yields ~ms then returns
+   the next pending key, or nil if no key was queued during the wait. Safe in
+   both native and WASM — `term/read-key` is only entered when a key is already
+   pending, so it returns immediately without blocking the runtime."
   [ms]
-  (let [armed (atom nil)]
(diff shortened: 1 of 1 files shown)


mparrett:
@nnunley FYI this was debugged and crafted fairly autonomously, and I understand it undoes some of your recent work. Happy to discuss and change the approach if I'm missing context.

nnunley (approved):
No worries on overwriting what I was trying to put together. I've been moving a bit less carefully than I realized, so its better to have this fixed and working. That said, it would be nice to have fast communication channels to bounce ideas around.

mparrett:
Re-pinging with proper context after reading #55 and #56 — I'd missed those when I opened this.

This PR is essentially a tactical implementation of one slice of the `input-source` contract you spec'd in #56 (`key-pending?` non-blocking probe, fall through to `read-key` only when armed). The deploy-regression bisect landed on #51 and I went straight to the smallest revert; only after did I see that #56's PR2 was already planned to supersede #51 inside the larger `xsofy/ui.lg` event-loop rollout, for exactly the WASM-`Atomics.wait` reason I describe above.

So the question is really about sequencing, not approach:

- If PR1/PR2 of #56 is imminent, I'm happy to close this and let the event-loop work absorb the fix cleanly — the `run-loop` model handles the `{:world :ui}` split way better than my point fix does, and `make-key-poller` would presumably get replaced wholesale by the source contract anyway.
- If #56 is further out, this unblocks the live deploy today with a change that's small enough to delete when the proper surface lands. The diff is `xsofy/screenfx.lg` only and doesn't touch `dispatch`/`game-loop`/seed handling, so it shouldn't conflict with the event-loop PRs structurally.

Defer to you on which makes sense. No attachment to landing this one specifically — happy to wait if the larger work is close.

PR #62, opened 2026-06-05
Reproducible runs: headless-safe render + seed-derived title/quest/scheme

nnunley:
## Reproducible runs (1/5)

Makes a run fully reproducible from its seed, and hardens headless rendering.

- **`render/term-dims`** — headless-safe terminal size (falls back to game dims when `term/size` is nil: CI/tests, or the brief resize gap). Removes nil-destructure crashes and a redraw-glitch source; routes all render entries through it.
- **`xsofy.seed/boot-seed`** — run seed from `*boot-seed*` (tests) / `XSOFY_SEED` env / random fallback.
- **Seed-derived title/quest/scheme** — `generate-title`/`generate-quest`/`pick-scheme` thread the deterministic RNG (`xsofy.det`) so a fixed seed reproduces the *entire* run (title, quest, scheme, world). let-go has no global RNG seeding, so these previously used the unseedable global `rand-int`.

Foundation for the `?seed=` / replay work in the rest of the stack.

### Stack (bottom→top)
1. reproducible-runs
2. repro-test-infra
3. native-xxh3
4. browser-gate
5. seed-and-replay

Tests: full let-go suite green.

### DIFF
--- a/main.lg
+++ b/main.lg
@@ -14,7 +14,8 @@
             [xsofy.title :as title]
             [xsofy.screenfx :as sfx]
             [xsofy.debug :as dbg]
-            [xsofy.dispatch :as dispatch]))
+            [xsofy.dispatch :as dispatch]
+            [xsofy.seed :as seed]))
 
 
 (defn init-terminal []
@@ -359,10 +360,12 @@
       :else world)))
 
 (defn new-game []
-  (let [{:keys [title scheme]} (title/show-title-screen)
-        quest (title/generate-quest)
+  (let [sd  (seed/boot-seed)
+        rng {:seed sd}
+        {:keys [title scheme rng]} (title/show-title-screen rng)
+        [quest rng] (title/generate-quest rng)
         _ (title/show-descending-screen 1 scheme)]
-    (-> (world/make-world 79 30 (rand-int 1000000000))
+    (-> (world/make-world 79 30 sd)
         (assoc :title title
                :scheme scheme
                :quest-item (:item quest))
--- a/xsofy/render.lg
+++ b/xsofy/render.lg
@@ -17,6 +17,12 @@
 
 (def dim-color u/scale)
 
+;; Terminal dimensions, headless-safe. term/size is nil with no TTY (CI/tests)
+;; and momentarily nil during a resize — both yielded a nil destructure and a
+;; blank/garbled redraw. Fall back to the game's nominal size.
+(defn term-dims []
+  (or (term/size) [ui/game-width ui/game-height]))
+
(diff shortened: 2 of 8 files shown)


mparrett (commented):
Reviewed this stack end-to-end (#62→#66) — see PR #66 for the cumulative review; separate per-PR comments on each. Posting in topological order so cross-references resolve as you read down.

The rng-threading rewrite in `xsofy/title.lg` is clean — passing `[result rng]` pairs through `generate-title` / `pick-scheme` / `generate-quest` keeps the pick order explicit and reproducible, and capturing `seed` in `show-title-screen` before any picks advance the rng is exactly right. Same for the `render/term-dims` helper. Two notes:

**1. `seed/boot-seed` accepts whatever `read-string` returns.**

```clojure
(let [s (os/getenv "XSOFY_SEED")]
  (when (and s (> (count s) 0)) (read-string s)))
```

A malformed `XSOFY_SEED` (e.g. `?seed=foo`, `?seed=12.5`, `?seed=`) reads as a symbol / float / nil and the failure surfaces somewhere later in the seed chain rather than at the boundary. One-line guard:

```clojure
(let [s (os/getenv "XSOFY_SEED")]
  (when (and s (> (count s) 0))
    (let [n (read-string s)]
      (when (integer? n) n))))
```

Same shape would apply to `boot-replay` in #66 (string-only is enough there — a base64 alphabet check would be even better).

**2. Rebase will collide with `nooga/xsofy#61`.**

The `make-key-poller` rewrite in `xsofy/screenfx.lg` is functionally identical to commit `96174c4` ("fix(title): use key-pending? instead of goroutine+alts! for WASM-safe input poll (#61)"), which is already on `main`. The rebase will either auto-resolve (likely) or produce a trivial conflict. The fixme'd animation test in #65 references #61 explicitly, so I assume you're aware — flagging so you can decide whether to drop your commit on rebase vs. keep it and let git merge them.

No blockers in this PR; the issues above are both 1–2 line fixes. Good design otherwise.

nnunley:
Thanks for the careful pass — addressed in the force-pushed update:

- **`boot-seed` guard** — now validates `read-string` returns an integer, else falls through to random, so `?seed=foo` / `12.5` / empty no longer surface later in the seed chain. `boot-replay` stays string-only as you suggested; the base64 alphabet check lives in `replay/decode` now (see #66 R5).
- **#61 rebase** — confirmed: the `make-key-poller` rewrite is functionally `96174c4` (already on `main`). Kept as-is; merges cleanly.

nooga (approved):

Issue #44, opened 2026-05-30
Need game release version number in web version for bug reporting


metalivedev:
The current version of `xsofy` no longer works on
* DuckDuckGo browser version 0.157.6 or 
* Google Chrome Version 148.0.7778.217 (Official Build) (64-bit)

Loading the page `https://nooga.github.io/xsofy/`
shows "loading" then 
```
[1;31merror:[0m syscall/js: call of Value.Int on undefined

[1mstack trace:[0m
  at require (<unknown>)

[program exited]
```

### Meta Bug/Feature Request:
It would be helpful if the `xsofy` WASM executable reported some kind of build id/version number, so that we could know which release is exhibiting the problem and include it in our bug reports.

Apologies for not diving deeper into this -- I don't have a WASM debugging environment set up.

mparrett:
@metalivedev Thanks for reporting this. Agreed we need a better way to track. At a glance looks like the published xsofy might not be using the required version of let-go. I've touched this recently and will take a look later tonight when I'm back at it.

mparrett:
Opened [let-go#132](https://github.com/nooga/let-go/issues/132) to discuss.

mparrett:
In flight in #137: build provenance is now surfaced in the shell — a version field (nearest v* tag via git describe, so a tagged Pages build reports exactly its release) shown in the ⚙ config pane with the full sha/date/lg provenance, plus a dim "xsofy v0.0.2"-style stand-in in the title bar until the game announces itself. Web bug reports get a version without opening devtools once #137 lands.

PR #63, opened 2026-06-05
Repro test infra: Layer-1 runtime smoke + ?seed= browser bridge

nnunley:
## Repro test infra (2/5)

- **Layer-1 runtime smoke** (`xsofy/test/smoke.lg` + `check_smoke.py`) — drives the real render/dispatch paths headlessly and asserts the title *animates* (frame 0 ≠ frame 50, a frozen-title guard) and the game renders + survives scripted turns. Runs fast on every PR, no browser. (`term/write` emits ANSI to stdout even off-TTY, so a Python comparator checks the captured frames — they can't be captured inside a let-go test.)
- **`tools/patch_wasm_seed.lg`** — bridges `?seed=` → the worker's `go.env.XSOFY_SEED` in the generated `index.html` (the upstream wrapper passes no env/argv to the wasm). Ships reproducible runs: `…/xsofy/?seed=12345` replays a run — serves #44. Idempotent + fail-closed, mirroring `patch_wasm_coi.lg`.

Depends on #62 (boot-seed / seed-derived runs).

### Stack (bottom→top)
1. reproducible-runs (#62)
2. repro-test-infra
3. native-xxh3
4. browser-gate
5. seed-and-replay

Tests: full let-go suite green; Layer-1 smoke OK; patch verified idempotent + composes with the COI patch.


mparrett (commented):
No requested changes in this PR — flagging one forward-looking concern so it lands cleanly when #66 extends the patcher.

The `tools/patch_wasm_seed.lg` design is solid: exact-string anchors, fail-closed `abort!` on missing/duplicate anchors, anchored on the unique `// Run Go WASM` comment to avoid main-thread/worker ambiguity, idempotency guard for re-runs. Same shape as `patch_wasm_coi.lg`.

The `xsofy/test/smoke.lg` + `check_smoke.py` driver is also a good shape — the `@@T0@@/@@T50@@/@@GAME/@@SMOKE-OK@@` sentinel stream gives the Python comparator both content checks (T0 ≠ T50, segments non-empty) and a "did it crash mid-render" check (missing sentinel = a render path threw and cut stdout short).

**One forward-looking note** — and worth fixing when you rebase #66 rather than here:

The idempotency guard checks `(str/includes? html "XSOFY_SEED")`. After #66 extends the same patcher to also inject `XSOFY_REPLAY`, a bundle that was patched by *this* version of `patch_wasm_seed.lg` and not rebuilt will be treated as complete — `XSOFY_REPLAY` will never get injected even after the patcher itself is updated. This bites partial rebuilds and any CI that caches `dist/` across stack states.

Two fixes for #66 to consider:

```clojure
;; (a) gate on both current outputs
(if (and (str/includes? html "XSOFY_SEED") (str/includes? html "XSOFY_REPLAY"))
  ...)

;; (b) inject an owned versioned sentinel the patcher controls
;;     e.g. write "<!-- xsofy-url-bridge:v2 -->" as part of the patch,
;;     guard on the matching sentinel string. Each patcher version
;;     owns its own idempotency state.
```

(b) is more durable for future bridge extensions; (a) is the smaller change.

Otherwise this PR looks good as-is.

nnunley:
Good forward-looking catch. Fixed in **#66** (where the patcher gains `XSOFY_REPLAY` injection): the idempotency guard now gates on **both** `XSOFY_SEED` and `XSOFY_REPLAY` (option (a)), so a bundle patched by the seed-only version and not rebuilt still receives the replay bridge on a partial rebuild / cached `dist/`.

nooga (approved):

PR #64, opened 2026-06-05
Swap to native xxh3 (value-preserving; −1486 lines)

nnunley:
## Native xxh3 (3/5)

`xsofy.hash` reimplemented xxh3-64 in ~1,486 lines of pure Lisp. let-go 1.9.0 ships a native `xxh3` namespace (interop to github.com/zeebo/xxh3) that is **bit-identical** (verified at seed 0 and nonzero). Swap the primitive, keep the `encode-salt` / `combine` / `with-folded-salt` API, delete the pure-Lisp mixer + the test-only streaming path.

**Value-preserving** — existing worlds, replays, and seeds stay byte-identical (no reseed). The full determinism suite (make-world reproducibility, replay equality, det/world-seed/dispatch/dag) passing is the proof; a pinned `native-xxh3-preserves-values` test guards the goldens.

Depends on #62.

### Stack (bottom→top)
1. reproducible-runs (#62)
2. repro-test-infra
3. native-xxh3
4. browser-gate
5. seed-and-replay

Tests: full let-go suite green (combine goldens + seeded title unchanged through the swap).


mparrett (commented):
The xxh3 swap design itself is good — same public surface (`xxh3-64`, `combine`), the macro-folding pipeline preserved, native interop via `xxh3/HashSeed` is in let-go fork tag `v1.9.0` so the published-upstream floor is OK. Going native also eliminates the WASM-vs-native divergence risk inside a hand-ported mixer.

Two blockers and one coverage concern though:

---

**🔴 Blocker 1 — `xsofy/test/hash_test.lg` fails to compile.**

The PR deletes `u64*-128` from `xsofy/hash.lg` but leaves `u64*-128-is-commutative` at `xsofy/test/hash_test.lg:143-145`. Verified locally:

```bash
$ lg xsofy/test/hash_test.lg
error: Can't resolve h/u64*-128 in this context
  --> xsofy/test/hash_test.lg:145:6
     |
 145 |   (= (h/u64*-128 a b) (h/u64*-128 b a)))
     |      ^^^
```

The native wrapper has no equivalent to `u64*-128` (the 128-bit product is hidden inside `xxh3/HashSeed`), so the property should just be deleted in this PR.

---

**🔴 Blocker 2 — `xsofy/test/run.lg` masks the load failure; CI will be green.**

Even with Blocker 1 active, the aggregate runner exits 0:

```bash
$ lg xsofy/test/run.lg; echo $?
...
[the hash_test.lg compile error prints here]
...
Finished running tests. Tests: 224 Pass: 2115 Fail: 0 Error: 0
0
```

CI's "Run tests" step uses this runner, so the workflow can stay green while a required test namespace silently failed to load. This is the kind of bug that hides every subsequent regression, not just this one — it's why Blocker 1 hasn't already been caught.

Two paths:

- **Cheap:** add per-namespace direct compile gates to CI (`lg -c /tmp/x.lgb xsofy/test/hash_test.lg` for each test ns). Any compile failure exits non-zero.
- **Right:** make the aggregate runner propagate load failures to the exit code. May need a small change on the let-go fork side to surface them.

This isn't strictly "your" bug — the masking has been latent — but landing #64 surfaces it, so it's the right PR to fix it on (or at minimum add the CI gate as a precondition).

---

**Coverage — the "bit-identical" parity claim is undertested.**

After fixing Blocker 1, the retained `hash_test.lg` covers only `xxh3-64("hello", 0)` plus two `combine` cases. xxh3-64's internal length-regime branches are at 1–3, 4–8, 9–16, 17–128, 129–240, ≥241 bytes — none exercised. Seeds: only 0 and 12345. High-bit-set bytes (≥128): not tested.

Suggested parity matrix to restore in `hash_test.lg`:

- Lengths {0, 1, 2, 3, 4, 8, 9, 16, 17, 32, 33, 64, 65, 96, 97, 128, 129, 240, 241, 256, 1024}
- Seeds {0, 42, -1, 0x8000000000000000}
- Byte sets including 0, 127, 128, 255
- Macro-folded `with-folded-salt` path with a fully-literal vector

Cross-check against `xxhsum -H3 --seed=N` or Python `xxhash.xxh3_64_intdigest(b, seed=N)` for the golden values. Then "bit-identical" is actually proven.

---

**Recommendation:** B1 + B2 land before merge; restore parity matrix in the same or follow-up PR. Design itself is good.

nnunley:
Both blockers + the coverage gap are addressed in the force-pushed update:

- **🔴 B1** — deleted the dead `u64*-128-is-commutative` property (the fn is gone; `u64+`/`u64*` remain, so the other two props stay).
- **🔴 B2** — added a per-namespace compile gate to CI (`test.yml`): `let-go -c` over every `*_test.lg` / `*_prop.lg` before the aggregate runner, so a masked load failure now fails the job. (Cheap path — runner-side propagation would need a let-go change.) Fixing B1 also un-masked the suite: **test count 224 → 244**.
- **Parity matrix** — restored, data-driven over your length regimes `{0,1,2,3,4,8,9,16,17,32,33,64,65,96,97,128,129,240,241,256,1024}` × seeds `{0, 42, -1, i64-min}` × a 0..255 byte pattern. Goldens are pinned from the native impl as a regression guard, **and** the seed-0 ASCII lengths are cross-checked against `xxh3sum` (e.g. len-0 `2d06800538d394c2`, len-17 `714a04408e79b80f`) — so it's canonical XXH3, not just self-consistent.

nooga (approved):

PR #65, opened 2026-06-05
Headless-browser e2e gate (boot + seeded regression)

nnunley:
## Browser gate (4/5)

A `@playwright/test` suite under `tests/e2e/` (separate from the existing `tests/browser/smoke.mjs`, which it complements):
- **Worker-mode boot** — cross-origin isolated, renders the title, console-clean.
- **Seeded regression** — `/?seed=12345` → the seed-derived title "Chasms of Infinity". This single assertion proves **both** that `?seed=` reached `os/getenv` **and** that wasm-xxh3 == native-xxh3 (a divergence in either changes the title).

Reads xterm via `.xterm-rows` textContent (the `#terminal` innerText is empty for the canvas renderer). Wires the Layer-1 smoke + a `-w` wasm-build smoke into `test.yml` (PRs), and the `?seed=` patch + this e2e into `deploy-pages.yml`.

Depends on #62, the ?seed= bridge, and native xxh3.

### Stack (bottom→top)
1. reproducible-runs (#62)
2. repro-test-infra
3. native-xxh3
4. browser-gate
5. seed-and-replay

Tests: e2e verified green locally against a built bundle.

### DIFF
--- a/.github/workflows/deploy-pages.yml
+++ b/.github/workflows/deploy-pages.yml
@@ -37,6 +37,9 @@ jobs:
       - name: Patch COI startup guard
         run: let-go tools/patch_wasm_coi.lg
 
+      - name: Bridge ?seed= for reproducible runs
+        run: let-go tools/patch_wasm_seed.lg
+
       - name: Set up Node (smoke gate)
         uses: actions/setup-node@v4
         with:
@@ -51,6 +54,17 @@ jobs:
       - name: Browser smoke (must boot to title screen)
         working-directory: tests/browser
         run: node smoke.mjs --dir ../../dist
+
+      # Seeded determinism regression: boots /?seed=12345 and asserts the
+      # seed-derived title — proving the ?seed= bridge AND wasm==native xxh3.
+      # Non-blocking until validated by a real CI run; then drop continue-on-error.
+      - name: E2E seeded regression (@playwright/test)
+        continue-on-error: true
+        working-directory: tests/e2e
+        run: |
+          npm install --no-audit --no-fund
+          npx playwright install chromium
+          DIST="$GITHUB_WORKSPACE/dist" npx playwright test
 
       - name: Upload smoke failure artifacts
         if: failure()
--- a/.github/workflows/deploy-pages.yml
+++ b/.github/workflows/deploy-pages.yml
@@ -57,9 +57,9 @@ jobs:
 
       # Seeded determinism regression: boots /?seed=12345 and asserts the
       # seed-derived title — proving the ?seed= bridge AND wasm==native xxh3.
-      # Non-blocking until validated by a real CI run; then drop continue-on-error.
+      # (The title-animation gate is the separate `smoke.mjs` step above /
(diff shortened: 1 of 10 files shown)


mparrett (commented):
Header-served COI as the CI baseline is the right call — the SW bootstrap path has been historically flaky and is host-specific (see the 2026-05-17 COI saga in our project notes). Worth documenting that gap explicitly in the spec file or README so it's not silently uncovered. Otherwise the spec shape — boot smoke + regression + animation gate, with the regression test asserting seed 12345 → "Chasms of Infinity" — is exactly the right kind of end-to-end backstop for this stack.

Three smaller items:

**1. Console-clean gate only fires in the first boot test.**

`expect(errors, errors.join('\n')).toEqual([])` is set up in `boot-smoke.spec.js` but not in `regression.spec.js` or the fixme'd animation test. So a regression that introduces a `console.error` during animation/replay passes silently as long as the terminal text assertion still resolves. Either factor it into a shared fixture or copy the listener+assertion into each test.

Also worth considering: is `expect(errors).toEqual([])` going to be too strict long-term? If you ever add intentional diagnostic logging via `console.error` (or let-go's `dbg/log` ever routes there for a panic), every test fails. An allowlist of known-noisy substrings might be more durable — but only add it when a real intentional case shows up.

**2. Port-release flake — fixed port 8123, `srv.kill()` not awaited.**

```js
} finally { srv.kill(); }
```

`spawn` + `kill` doesn't wait for the child to release the port before the test resolves. Back-to-back tests (or Playwright's `retries: 1`) can hit a port still in TIME_WAIT. Two options:

- await server shutdown (wrap `srv.kill()` + `await new Promise(r => srv.on('exit', r))`)
- rotate ports per worker via `PORT: String(8120 + test.info().workerIndex)` and pass it through to `static-server.js`

**3. `TITLE_FRAGMENT` comment is misleading.**

```js
const TITLE_FRAGMENT = 'Chasms';   // seed-independent: every title is "X of Y"; this seed yields "Chasms of Infinity"
```

The fragment IS seed-specific — different seeds give different titles. The comment reads as "this assertion would pass on any seed," which would be a bug. The test itself is fine; just fix the comment to say something like `// for seed 12345; change with the seed`.

---

**Recommendation:** all three are small. No blockers. Good gate; the points above harden it.

nnunley:
All three + the doc note:

- **console gate** — factored a `watchConsole(page)` helper and assert `errors == []` in the regression + animation tests, not just the first boot test.
- **port flake** — added `stopServer(srv)` that awaits the child's `exit` before resolving, so `8123` is released before the next test / retry.
- **`TITLE_FRAGMENT` comment** — corrected to say seed-SPECIFIC.
- **COI gap** — the SW-bootstrap path being intentionally uncovered is documented in `boot-smoke.spec.js`'s header comment.

Note: these spec files are last-shaped in #66 (the animation gate + `?replay=` test), so the fixes landed in #66 commits to stay conflict-free across the stack.

nooga (approved):

PR #66, opened 2026-06-05
Seed display + shareable replay codes (?replay=)

nnunley:
## Seed display + replay codes (5/5)

- **Show the run seed** on the title screen, in-game HUD, and death screen (it enables `?seed=` reproducibility — serves #44).
- **`xsofy.replay`** — pack a run (seed + `:action-log`) into a compact, URL-safe base64 **replay code** (self-describing action dictionary + RLE of the movement streaks; e.g. 200 `:up` → <40 chars). Round-trips and reconstructs the identical world via `dispatch/replay`.
- **Death screen** shows the replay code, wrapped.
- **Playback** — `XSOFY_REPLAY` / `?replay=CODE` decodes a code, rebuilds the run, and steps through the actions with a short delay so it plays out, then hands control back at the end-state. `?replay=` is bridged into `go.env` like `?seed=`.
- Enables the e2e title-animation gate now that nooga/xsofy#61 (WASM-safe input poll) is on main.

Depends on the rest of the stack.

### Stack (bottom→top)
1. reproducible-runs (#62)
2. repro-test-infra
3. native-xxh3
4. browser-gate
5. seed-and-replay

Tests: let-go suite green (incl. replay round-trip/reconstruction); e2e 4/4 green incl. `?replay=` playback.

### DIFF
--- a/main.lg
+++ b/main.lg
@@ -360,17 +360,51 @@
 
       :else world)))
 
+(defn assemble-world
+  "Build the starting world for `seed` and graft on its (seed-derived) title,
+   scheme, and quest. Title/scheme/quest are cosmetic — they don't touch the
+   gameplay seed chain — so this world replays identically to one made by
+   new-game from the same seed."
+  [seed title scheme quest]
+  (-> (world/make-world 79 30 seed)
+      (assoc :title title
+             :scheme scheme
+             :quest-item (:item quest))
+      (#(reduce world/log-msg % (:messages quest)))))
+
 (defn new-game []
   (let [sd  (seed/boot-seed)
         rng {:seed sd}
         {:keys [title scheme rng]} (title/show-title-screen rng)
         [quest rng] (title/generate-quest rng)
         _ (title/show-descending-screen 1 scheme)]
-    (-> (world/make-world 79 30 sd)
-        (assoc :title title
-               :scheme scheme
-               :quest-item (:item quest))
-        (#(reduce world/log-msg % (:messages quest))))))
+    (assemble-world sd title scheme quest)))
+
+(defn build-game
+  "Reconstruct a run's starting world from its seed, without the interactive
+   title screen (same pick order as new-game: title -> scheme -> quest)."
+  [seed]
+  (let [rng {:seed seed}
+        [title rng]  (title/generate-title rng)
+        [scheme rng] (title/pick-scheme rng)
+        [quest rng]  (title/generate-quest rng)]
+    (assemble-world seed title scheme quest)))
+
+(defn play-replay
+  "Decode a replay code, reconstruct the run, and step through its actions with
+   a short delay so the run plays out on screen. Returns the final world so the
+   caller can hand control back to the player at the recorded end-state."
+  [code]
+  (let [{:keys [seed actions]} (replay/decode code)
+        w0 (build-game seed)]
+    (render/render-full w0)
+    (sfx/pause! 400)
+    (reduce (fn [w action]
+              (let [w' (dispatch/dispatch w action)]
+                (render/render-full w')
+                (sfx/pause! 90)
+                w'))
+            w0 actions)))
 
 (defn game-loop
   ([world]
@@ -568,7 +602,8 @@
 (defn -main []
   (init-terminal)
   (try
-    (game-loop (new-game))
+    (let [code (seed/boot-replay)]
+      (game-loop (if code (play-replay code) (new-game))))
     (catch e
            (dbg/log "FATAL ERROR:" e)
            (dbg/dump-crash! e)
--- a/main.lg
+++ b/main.lg
@@ -15,7 +15,8 @@
             [xsofy.screenfx :as sfx]
             [xsofy.debug :as dbg]
             [xsofy.dispatch :as dispatch]
-            [xsofy.seed :as seed]))
+            [xsofy.seed :as seed]
+            [xsofy.replay :as replay]))
 
 
 (defn init-terminal []
@@ -434,6 +435,7 @@
          :death
          (let [[tw th] (term/size)
                runes (render/death-runes tw th)  ;; scatter ONCE — stable across frames
+               replay-code (replay/encode-run world)  ;; compute ONCE — encode is O(n) per call
                poll! (sfx/make-key-poller 120)  ;; calmer death-screen pulse (was 80ms)
                _ (sfx/clear-screen tw th)  ;; clear ONCE, not per-frame — was the flicker source
                result (loop [scroll 0 frame 0]
@@ -441,7 +443,7 @@
                         ;; pulsing while we wait for a key instead of freezing.
                         (let [[k nf] (sfx/drive-until-key
                                        frame
-                                       (fn [f] (render/render-death-screen world runes scroll f))
+                                       (fn [f] (render/render-death-screen world runes scroll f replay-code))
                                        poll!)]
                           (cond
                             (or (= k "n") (= k "\r") (= k "\n") (= k " "))
--- a/tests/e2e/boot-smoke.spec.js
+++ b/tests/e2e/boot-smoke.spec.js
@@ -28,6 +28,19 @@ async function waitForServer(port = 8123, timeout = 5000) {
   throw new Error(`Server did not start within ${timeout}ms`);
 }
 
+// Wait for the child to release the port before resolving — a bare srv.kill()
+// can leave 8123 in TIME_WAIT and flake the next test / retry.
+function stopServer(srv) {
+  return new Promise((resolve) => { srv.on('exit', () => resolve()); srv.kill(); });
+}
+
+function watchConsole(page) {
+  const errors = [];
+  page.on('console', (m) => { if (m.type() === 'error') errors.push(m.text()); });
+  page.on('pageerror', (e) => errors.push(String(e)));
+  return errors;
+}
+
 // xterm.js renders into .xterm-rows; its text is in textContent (NOT #terminal
 // innerText, which comes back empty).
 function termText(page) {
@@ -44,13 +57,11 @@ function waitForTermText(page, needle, timeout = 30000) {
 
 // The title text renders from frame 0, so it's present even if the animation
 // loop is frozen. (The "Press any key" subtitle is frame-30-gated — see below.)
-const TITLE_FRAGMENT = 'Chasms';   // seed-independent: every title is "X of Y"; this seed yields "Chasms of Infinity"
+const TITLE_FRAGMENT = 'Chasms';   // seed-SPECIFIC: fragment for the boot seed; different seeds → different titles
 
 test('worker-mode boot: cross-origin isolated, renders the title, console-clean', async ({ page }) => {
   const srv = serve();
-  const errors = [];
-  page.on('console', m => { if (m.type() === 'error') errors.push(m.text()); });
-  page.on('pageerror', e => errors.push(String(e)));
+  const errors = watchConsole(page);
   try {
     await waitForServer();
     await page.goto('/?seed=12345');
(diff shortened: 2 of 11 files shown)


mparrett (commented):
**Stack-level review.** I reviewed all 5 PRs as a stack (#62→#66) — per-PR comments on each cover issues scoped there. This comment covers PR #66 plus the cumulative concerns about the determinism claim, since that contract is what this PR establishes.

The capstone shape is right: seed display on title/HUD/death, self-describing replay codes, `?seed=`/`?replay=` URL bridges, browser playback. `build-game` reconstructing the run without the interactive title screen (using the same pick order as `new-game`) is a clean architectural move. The wire format choice — version byte + origin seed + self-contained keyword dictionary + RLE + URL-safe base64 — is a good v1 concept.

**But the implementation overclaims what the action log actually captures, and the wire format trusts too much.** The findings below are why I'd hold this PR until D1 + D2 + R5 land.

---

## 🔴 D1 — Replay overclaims: inventory/menu/inscribe/hazard handlers bypass dispatch

`xsofy/replay.lg`'s docstring says "A run is fully determined by `(seed, action-log)`." That holds for the subset of play that flows through `dispatch/dispatch`. It does *not* hold for actual gameplay. Verified call sites in `main.lg` that mutate the world directly without enqueueing to `:action-log`:

| Line | Handler | Bypasses dispatch via |
|---|---|---|
| `main.lg:59` | `inv-equip` | `world/equip-item` |
| `main.lg:87` | `inv-drop` | `world/drop-item` |
| `main.lg:107` | `inv-use` | `world/use-item` |
| `main.lg:135` | `inscribe-input` | rune inscription |
| `main.lg:226` | `execute-menu-action` | quick-menu use/drop/equip |
| `main.lg:512` | hazard yes/no | `world/confirm-hazard-step` / `cancel-hazard-step` |

So a real run with combat + pickups + a consumable use + a hazard confirmation cannot be reproduced from its replay code. The seed alone reproduces the opening state (title/quest/scheme/world generation); `:action-log` reproduces dispatch-routed turns; everything else drifts silently. The death screen's `replay code (paste into ?replay=):` UI implies completeness, which the contract doesn't deliver.

Two ways forward:

- **Narrow the claim (pragmatic).** Update `xsofy/replay.lg`'s docstring + the death-screen UI label to "keyword-dispatch gameplay only — runs involving inventory/menu/inscribe/hazard interactions cannot be fully reproduced from the replay code." Same seed still reproduces the opening state, which is meaningful on its own. Lets this stack land.
- **Widen the contract (right, follow-up PR).** Route the bypass paths through `dispatch/dispatch` with parameterized action maps. Dispatch already supports them (`xsofy/dispatch.lg:30` documents `{:type :use-item :id :potion-7}`). Requires extending the wire format to carry a per-map payload block; meaningful work, deserves its own PR.

For this stack, I'd go with **narrow**. It's a one-line docstring + UI change and stops the contract from overclaiming.

---

## 🔴 D2 — `encode-run` will throw on map actions

`xsofy/dispatch.lg:75` unconditionally `(update w' :action-log conj action)`, and per the docstring the action can be a map: `{:type :use-item :id :potion-7}`. But `xsofy/replay.lg`:

```clojure
(defn- name-bytes [kw]
  (mapv int (name kw)))   ;; action keyword names are ASCII
```

`(name {...})` throws — `name` doesn't accept maps. The moment any caller routes a map action through dispatch (which is a documented feature), `encode-run` fails at the death screen with a runtime error.

Cheap defense in this PR:

```clojure
(defn encode-run [world]
  (let [actions (or (:action-log world) [])]
    (assert (every? keyword? actions)
            "replay/encode-run: action-log contains a non-keyword action; \
             replay codes currently support keyword-only logs")
    (encode (:seed-input world) actions)))
```

Or filter + log:

```clojure
(let [actions (filterv keyword? (:action-log world))]
  ...)
```

The real fix lands with D1's widening — the format needs a per-map payload block. For this PR, assert-on-encode is the right defense.

---

## 🟡 R5 — Replay format trusts the user-pasted input

All findings below verified by running `replay/encode` and `replay/decode` against edge inputs in a let-go REPL. None are catastrophic alone; collectively they should be tightened before the `?replay=` URL becomes a public surface.

**Dictionary cap is documented but not enforced.** `encode` writes `(count dict)` into a one-byte field; dict indices are also one byte. Tested: 256 distinct actions encode succeeds, decode fails with `nth index out of bounds`. Assert `< 256` on encode for dict size, name length, run index, and dict index (on decode too).

**`name-bytes` isn't UTF-8 despite the docstring claim.** `(mapv int (name kw))` is codepoint bytes. ASCII action keywords round-trip; an action keyword containing a non-ASCII char (e.g. `:é`, an emoji) produces a malformed code that crashes on decode. Either explicitly assert ASCII-only on encode + document, or use real UTF-8 encoding for the dictionary entries.

**`b64-decode` doesn't validate alphabet.** Pasting `?replay=!!!!` produces a low-level `bit-shift-left expected Int` rather than a replay-format error. User-pasted bad codes should produce one controlled failure path. Suggest a single `validate-code` step at the top of `decode` that checks alphabet membership + length plausibility before doing the byte work.

**`read-varint` and `decode` don't bound trailing bytes or huge varints.** A truncated code crashes through `nth` eventually; a malicious code with a giant varint run length allocates a large action vector before any sanity check. Bound the varint to a reasonable max (say, `< 100000`).

**Version-mismatch is exact-match only.** That's the right behavior for v1, but the docstring should say "v1-exact; future versions will reject v1 codes by default." Otherwise the next person to bump VERSION assumes migration is intended.

---

## 🟡 R4 — Renamed actions silently no-op (related to D1)

If a future action keyword gets renamed (`:up` → `:north`), old replay codes decode the old keyword. `dispatch/dispatch` treats unknown non-UI keywords as replay actions: folds them into the seed via `seed/advance`, appends to `:action-log`, then `world/update-world` falls through to a no-op for the unknown action. So seed advances but world doesn't mutate — silent drift, the run "plays back" but produces the wrong final state.

Two fixes, both small:

1. In `xsofy/replay.lg` docstring, add: *Action keywords are wire-format ABI. Renaming an action keyword breaks all existing replay codes. The format does not migrate.*
2. In `xsofy/dispatch.lg`, add an explicit unknown-action handler that aborts replay (or at minimum logs a warning and doesn't advance the seed).

---

## 🟡 R3 — Round-trip test under-covers

`replay-code-reconstructs-the-same-world` checks `:seed`, `:action-log`, `:terrain`. Drift in `:entities`, `:fov`, `:turn`, `:log`, floor cache, item state, or pending-screen state slips through. Suggest expanding:

```clojure
(is (= (:entities live) (:entities rebuilt)))
(is (= (:fov live)      (:fov rebuilt)))
(is (= (:turn live)     (:turn rebuilt)))
```

Or property-test: random seed, random valid-for-state action sequence, assert full-world equality post-replay.

---

## 🟡 R2 (fix here) — Re-key the patcher idempotency guard

`tools/patch_wasm_seed.lg:51` checks `(str/includes? html "XSOFY_SEED")`. This PR extends the patcher to also inject `XSOFY_REPLAY`, but the guard still only checks for `XSOFY_SEED`. A bundle previously patched by the #63 version (seed only, no replay) and not rebuilt will be treated as complete and never receive the replay bridge. Affects partial rebuilds + cached CI `dist/`.

Two fixes:

```clojure
;; (a) gate on both current outputs
(if (and (str/includes? html "XSOFY_SEED") (str/includes? html "XSOFY_REPLAY"))
  (println "...") ...)

;; (b) inject an owned versioned sentinel (more durable for future extensions)
;;     "<!-- xsofy-url-bridge:v2 -->" written by the patcher; guard on the
;;     matching sentinel string. Each patcher version owns its own idempotency state.
```

---

## 🟢 Smaller items

**Print `?replay=<code>` directly on the death screen.** The seed line is symmetric (`?seed=N to replay`); the replay block prints `replay code (paste into ?replay=):` followed by the bare code. Asymmetric — print `?replay=<code>` so a player can copy a single share-ready URL fragment in one swipe. The wrap logic stays the same; just prefix the first line of the wrapped block.

**Phone-portrait viewport.** `cw = max 24 (- tw 4)`; on a 79-wide portrait viewport a long log produces a tall wrapped block that can occlude the message log. Worth a real-device test (or maybe even a snapshot in tests/e2e/). If it's an issue, consider truncating with `...` and noting that the seed alone still reproduces the opening state.

---

## Recommendation

**Hold this PR until D1 + D2 land** (D1 can be the narrow-claim route — small change). R5/R4/R3/R2 should also land before merge; they're all bounded fixes inside this PR.

Architecture itself stands up: native hash interop (#64) + deterministic seed plumbing (#62) + URL bridges (#63/#66) + browser gate (#65) are the right pieces for this story. The work above is sharpening the implementation to match what the design promises.

nnunley:
Held the right things — D1 + D2 + R5 are in. Resolutions:

- **🔴 D1** — went **narrow** as recommended: `replay.lg`'s docstring now scopes the claim to keyword-dispatch gameplay (seed reproduces the opening state; inventory/menu/inscribe/hazard interactions are not fully reproducible from the code). The **widen** path is real follow-up — I already have most of it (use/equip/unequip/drop/enchant → parameterized dispatch actions, with a `(seed, action-log)` replay property test) on a separate branch; it'll be its own PR.
- **🔴 D2** — `encode-run` now asserts keyword-only and throws a clear error rather than crashing in `name-bytes`.
- **🟡 R5** — dict size capped (`< 256`) on encode; `name-bytes` asserts ASCII; `b64-decode` rejects out-of-alphabet chars with one controlled error; `decode`/varint bound run-count, run-length, and dict-index; version note added (v1-exact).
- **🟡 R4** — ABI note added to the docstring. The dispatch-side unknown-action guard targets `dispatch.lg` (authored in #53, outside this stack) — deferred to a dispatch-layer change.
- **🟡 R3** — round-trip test now also asserts `:entities` / `:fov` / `:turn`.
- **🟡 R2** — patcher guard re-keyed (see #63 reply).

Thanks for the thorough review — the B1/B2 pair in particular surfaced that the runner was silently masking load failures, which was hiding more than just this stack.

mparrett (commented):
Fast and thorough turnaround — thanks. Pulled the new stack tip locally and exercised the gates:

- `lg xsofy/test/hash_test.lg` → exit 0, no `Can't resolve` (B1 ✅)
- `lg xsofy/test/run.lg` → **`Tests: 244 Pass: 2228 Fail: 0 Error: 0`** (was 224 pre-fix; B1 + B2 ✅)
- Ran the `.github/workflows/test.yml` per-namespace compile loop manually: all 15 test ns OK on green code, and a synthetic broken ns (`(this-symbol-does-not-exist)` in a temp `_test.lg`) correctly produces `::error::… failed to compile` + non-zero exit — so B2's gate actually surfaces what the aggregate was hiding (verified ✅)
- Layer-1 smoke: `OK: title animates; game renders; no render threw`

Good design choices worth calling out:

- **xxh3 canonical anchor** — pinning goldens from the native impl alone would only prove self-consistency; cross-checking len 0/1/17 against `xxh3sum` makes the parity matrix a real canonical-XXH3 proof. Right move.
- **D1 narrow + widen on a separate branch** — clean split. Looking forward to the dispatch-routed inventory/menu/inscribe/hazard PR when it lands.
- **R5 structured `ex-info` errors** end-to-end — invalid `?replay=` now produces one controlled failure path, not random `bit-shift-left expected Int` deep in the byte math.
- **Patcher gated on both `XSOFY_SEED` + `XSOFY_REPLAY`** + the renamed "url-bridged" message — clean. One operational note for anyone reading this: existing seed-only `dist/` bundles need to be rebuilt; the new patcher correctly refuses to upgrade them in place.

Three items still open (none blockers from where I sit):

- **R4 — dispatch-side unknown-action guard.** Agreed it's `dispatch.lg`'s problem, not this stack's. The ABI note in `replay.lg`'s docstring is the right thing to land here; the actual guard belongs with the widen-the-contract PR.
- **R11 — death-screen `?replay=<code>` line.** Cheap polish: the seed line is symmetric (`?seed=N to replay`); printing `?replay=<code>` as the first line of the wrapped block (in addition to or instead of `replay code (paste into ?replay=):`) lets a player copy a single share-ready URL fragment in one swipe.
- **R12 — long replay codes on phone-portrait.** Worth a real-device check before this surface goes public; if it occludes the message log, truncate with `…` and note that the seed alone reproduces the opening state.

LGTM from my side modulo R11/R12 cosmetics; deferring to @nooga for the merge call.

nooga (approved):

nooga:
Finally got to read all of this. Thanks for handling this @mparrett @nnunley! LGTM

nooga (approved):

PR #67, opened 2026-06-05
Widen the replay contract: route menu item-actions through dispatch

nnunley:
## What

Widens the deterministic-replay contract so **menu-driven item actions** flow through `xsofy.dispatch` and are folded into the seed + `:action-log`, instead of mutating the world directly from `main`. Stacks on #66.

Covered actions: `use` (already routed), **`equip`**, **`unequip`**, **`drop`**, **`enchant`**.

## Why

`xsofy/replay.lg` currently documents the gap:

> Runs that involve inventory, menu, inscribe, or hazard-confirm interactions are NOT fully reproducible from the code today — those handlers mutate the world outside dispatch (widening that contract is a follow-up).

Item actions taken from the quick-menu/inventory bypassed dispatch, so they never entered the seed/log and didn't replay. This closes that for equip/unequip/drop/enchant.

## How

- **`world/update-world`** gains typed handlers for parameterized actions: `{:type :equip|:unequip|:drop :id …}` (new `equip-action`/`unequip-action`/`drop-action`) and `{:type :enchant :id target :scroll-id scroll}` (reuses existing `enchant-item`). The cursed/equipped guards + messages move from `main` into `world`.
- **`main`** — `execute-menu-action` and `inv-equip/unequip/drop/use` now just *assemble and dispatch* the typed action and manage screen state; the rules live in `world`.
- Since `:equip/:unequip/:drop/:use-item/:enchant` are **not** UI actions, `dispatch` folds + logs them automatically, so they replay bit-identically.
- **Enchant (two-step)** stays replay-consistent: using the scroll is a dispatched `:use-item` that opens the *pure-UI* target submenu without consuming; only the final `{:type :enchant …}` mutates. Cancelling leaves the scroll unconsumed in both live and replay.

## Tests

`xsofy/test/menu_dispatch_test.lg`: each action mutates correctly, is logged, advances the seed, and `dispatch/replay` reproduces `:entities` exactly; plus a dispatch-purity + consume/enchant test for the two-step flow.

**249 tests / 2246 assertions / 0 fail / 0 error.** `main.lg` compiles; smoke green.

## Scope / follow-up

This widens item actions. The replay docstring also lists **`inscribe-input`** and the **hazard-confirm** handler as bypassing dispatch — those are **not** covered here and remain a follow-up to fully honor the contract. Relates to #56 (PR3b determinism core).

### DIFF
--- a/main.lg
+++ b/main.lg
@@ -422,7 +422,11 @@
          :death
          (let [[tw th] (term/size)
                runes (render/death-runes tw th)  ;; scatter ONCE — stable across frames
-               replay-code (replay/encode-run world)  ;; compute ONCE — encode is O(n) per call
+               ;; compute ONCE — encode is O(n) per call. encode-run-safe yields
+               ;; nil for runs whose log has map actions (item use/equip/etc via
+               ;; dispatch); the death screen then shows the seed alone, instead
+               ;; of crashing the game on encode-run's keyword-only guard.
+               replay-code (replay/encode-run-safe world)
                poll! (sfx/make-key-poller 120)  ;; calmer death-screen pulse (was 80ms)
                _ (sfx/clear-screen tw th)  ;; clear ONCE, not per-frame — was the flicker source
                result (loop [scroll 0 frame 0]
--- a/main.lg
+++ b/main.lg
@@ -218,7 +218,9 @@
                      :enchant
                      (let [scroll-id (:enchant-scroll-id world)]
                        (if scroll-id
-                         (-> (world/enchant-item world item-id scroll-id)
+                         ;; dispatch the assembled (target, scroll) action so the
+                         ;; enchant + scroll-consume fold into the seed + log
+                         (-> (dispatch/dispatch world {:type :enchant :id item-id :scroll-id scroll-id})
                              (dissoc :enchant-scroll-id))
                          world))
 
--- a/main.lg
+++ b/main.lg
@@ -56,65 +56,37 @@
     (when (and (>= sel 0) (< sel (count inv)))
       (nth inv sel nil))))
 
+;; The inventory-screen item actions, like the quick-menu, route world-mutating
+;; actions through dispatch (seed + :action-log) — guards/messages live in
+;; world's *-action handlers — and only manage the :inventory screen state here.
+
 (defn inv-equip [world]
   (if-let [item-id (sel-item-id world)]
-    (let [item (get-in world [:entities item-id])
-          slot (get-in item [:item :slot])
-          iname (get-item-name world item-id)]
-      (if (and slot (not= slot :consumable))
-        (-> (ent/equip-item world item-id :player)
-            (world/log-msg (str "You equip the " iname "."))
-            (assoc :screen :inventory))
-        (assoc world :screen :inventory)))
+    (-> (dispatch/dispatch world {:type :equip :id item-id})
+        (assoc :screen :inventory))
     (assoc world :screen :inventory)))
 
 (defn inv-unequip [world]
   (if-let [item-id (sel-item-id world)]
-    (let [item (get-in world [:entities item-id])
-          slot (get-in item [:item :slot])
-          iname (get-item-name world item-id)]
-      (if slot
-        (if (ent/cursed? world item-id)
-          (-> world
-              (world/log-msg (str "The " iname " is cursed! You can't remove it."))
-              (assoc :screen :inventory))
-          (-> (ent/unequip-item world :player slot)
-              (world/log-msg (str "You remove the " iname "."))
-              (assoc :screen :inventory)))
-        (assoc world :screen :inventory)))
+    (-> (dispatch/dispatch world {:type :unequip :id item-id})
+        (assoc :screen :inventory))
     (assoc world :screen :inventory)))
 
 (defn inv-drop [world]
   (if-let [item-id (sel-item-id world)]
-    (let [iname (get-item-name world item-id)
-          equipment (or (get-in world [:entities :player :equipment]) {})
-          slot (get-in world [:entities item-id :item :slot])
-          equipped-id (when slot (get equipment slot))
-          ;; can't drop equipped cursed items
-          is-equipped-cursed (and (= equipped-id item-id) (ent/cursed? world item-id))]
-      (if is-equipped-cursed
-        (-> world
-            (world/log-msg (str "The " iname " is cursed! You can't remove it."))
-            (assoc :screen :inventory))
-        (let [w (if (= equipped-id item-id)
-                  (ent/unequip-item world :player slot)
-                  world)]
-          (-> (ent/drop-item w item-id :player)
-              (world/log-msg (str "You drop the " iname "."))
-              (assoc :screen :inventory :inv-sel (max 0 (dec (inv-sel world))))))))
+    (-> (dispatch/dispatch world {:type :drop :id item-id})
+        (assoc :screen :inventory :inv-sel (max 0 (dec (inv-sel world)))))
     (assoc world :screen :inventory)))
 
 (defn inv-use [world]
   (if-let [item-id (sel-item-id world)]
-    (let [item (get-in world [:entities item-id])
-          slot (get-in item [:item :slot])]
-      (if (= slot :consumable)
-        (let [result (world/use-item world item-id)]
-          ;; if use-item opened a submenu (enchant), keep its screen
-          (if (:screen result)
-            (dissoc result :inv-sel)
-            (dissoc result :screen :inv-sel)))
-        (assoc world :screen :inventory)))
+    (if (= :consumable (get-in world [:entities item-id :item :slot]))
+      (let [result (dispatch/dispatch world {:type :use-item :id item-id})]
+        ;; if use-item opened a submenu (enchant), keep its screen
+        (if (:screen result)
+          (dissoc result :inv-sel)
+          (dissoc result :screen :inv-sel)))
+      (assoc world :screen :inventory))
     (assoc world :screen :inventory)))
 
 (defn inv-inscribe [world]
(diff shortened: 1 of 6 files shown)


mparrett (commented):
Read this on top of #66's tip. The refactor is clean — guards/messages move out of `main`'s six inline blocks into pure `(world, item-id) → world` handlers in `world.lg`, and `main` collapses to "assemble the typed action, dispatch, manage screen state." Reuses existing primitives (`ent/equip-item`, `ent/cursed?`, etc.) without modification. `dispatch.lg` doesn't need to change at all — the existing `(action-type action)` already routes maps, and the new types aren't in `ui-actions`, so they auto-qualify as replay actions. Tests assert the right contract: mutate ✓ log ✓ seed-advance ✓ `dispatch/replay` reproduces `:entities`. The enchant test honestly notes the scroll is spawn-based (not seed-derived) and asserts purity rather than replay-from-seed.

Findings, in rough priority order:

**1. The replay-CODE (wire format) contract is NOT widened — only in-process `dispatch/replay` is.** This is the most important framing point and the only reason I'd ask for a docstring update before merge.

`xsofy/replay/encode-run`'s `(every? keyword? actions)` guard from #66 still throws on `:action-log` entries like `{:type :equip :id <item-id>}`. So after this PR:

- ✅ `dispatch/replay 79 30 seed actions` reproduces these runs in-process (what your tests cover)
- ❌ `?replay=<code>` cannot share them — `encode-run` rejects the map actions

The narrative in `xsofy/replay.lg`'s docstring from #66 — *"Runs that involve inventory, menu, inscribe, or hazard-confirm interactions are NOT fully reproducible from the code today … widening that contract is a follow-up"* — is now stale for item actions. After #67, "in-process replay covers them; the v1 wire format does not yet." Two-line fix in `replay.lg`'s docstring would make the split explicit and avoid a reader-vs-implementation mismatch. (The wire-format extension itself can be follow-up; the docstring change is the bit that should land alongside #67's behavioral widening.)

**2. The enchant cancel path is claimed but not tested.** PR body: *"Cancelling leaves the scroll unconsumed in both live and replay."* No assertion exercises cancel.

Worth probing in a test: after `dispatch {:type :use-item :id scroll}` opens the submenu, what happens when the next dispatched action is movement (`:down`) instead of `:enchant`? The world state is `:screen :enchant-select` with `:enchant-scroll-id` set. Does `:down` route to player-move (because dispatch only sees the keyword, not the screen)? Or is the input layer screen-aware so `:down` never enters the log during submenu navigation?

The answer determines whether `[:use-item-scroll, :down, :down, …]` replays correctly when the live user pressed `:down` to navigate the submenu vs. to move. If the input layer routes screen-aware, the log will only contain `:use-item-scroll` followed by whatever real game action came after the cancel — and replay works. If not, there's a divergence. Either way, a one-action `[:use-item-scroll]` replay asserting `(:screen rep) = :enchant-select` and `(:entities rep) = (:entities live)` (scroll un-consumed in inventory) would be a useful pin.

**3. Minor behavioral regression: failed-drop `:inv-sel`.** Old `inv-drop`:

```clojure
(if is-equipped-cursed
  (-> world (log-msg "…cursed…") (assoc :screen :inventory))   ; no :inv-sel change
  (let [...] (... (assoc :screen :inventory :inv-sel (max 0 (dec (inv-sel world)))))))
```

New:

```clojure
(-> (dispatch/dispatch world {:type :drop :id item-id})
    (assoc :screen :inventory :inv-sel (max 0 (dec (inv-sel world)))))   ; always
```

So a failed cursed-equipped drop attempt now moves the inventory cursor by 1. Tiny UX wart. Two cheap fixes: (a) compare pre/post inventory count and only decrement if the count changed, or (b) compute `:inv-sel` from the dispatched world's inventory size rather than blindly decrementing.

**4. Failed-action seed-advance — consistent, but worth noting intentionally.** Cursed unequip, cursed-equipped drop, and equip-non-equippable all return the world unchanged from the `*-action` handlers, but `dispatch` has already advanced the seed and recorded the map in `:action-log` by the time control returns. Live and replay agree (so determinism is preserved), but "no-op" attempts still consume entropy. Fine for the deterministic contract; just flagging because it's a new surface area — every keypress-that-routes-to-a-replay-action now mutates the seed even when nothing visible happens.

If you'd want short-circuit semantics, dispatch could check `(identical? w' w)` post-`update-world` and skip the fold + log. That's a bigger design decision (touches every action), so out of scope here.

**5. `:enchant` is the first multi-id dispatchable action.** Dispatch folds `(action-type action)` only — per the comment in `dispatch.lg:72-73`, "encode-salt can't take a map; an action's params are reproduced via the recorded :action-log." So two enchants with different `(target, scroll)` pairs produce the same seed delta. Correct for determinism (the log carries the params), but means a "seed-only forensic replay" can't distinguish them. Not a bug; just the first PR to surface this. Worth a sentence in the enchant-action docstring noting that the params are log-only, not seed-folded.

**6. Happy-paths-only test coverage.** Solid for the claim. Nice-to-haves:

- Cursed unequip stays logged but doesn't mutate, and replays identically (would also pin the seed-advance-on-no-op behavior from #4 above).
- Mixed-action sequence `[:wait :equip :down :drop :wait]` reproduces `:entities` (the existing tests cover single actions in isolation; this would pin the contract under realistic interleaving).
- Equip-to-already-equipped-slot: does it auto-unequip and equip, or no-op? Either is fine but the behavior should be pinned.

## Recommendation

Mergeable subject to **#1 (docstring update)** and **#2 (cancel test or smoke note)**. **#3** is a small follow-up — same PR or separate, your call. The rest are observations.

For the follow-up that closes inscribe-input + hazard-confirm: the #67 pattern transfers cleanly. If you also extend the v1 wire format to carry a per-map payload block at the same time (or v2-it), that fully honors the original "widen the contract" framing across both layers — both `dispatch/replay` and `?replay=` would cover everything.

mparrett (commented):
After a closer trace, two of the findings in my review above are sharper than I framed them — both are real correctness issues, not just docs / open questions. Wanted to flag before you spin on the soft framing.

**Finding #1 is a runtime crash, not just stale docs.** `main.lg:425` (in the `:death` handler from #66) calls `replay/encode-run` unconditionally:

```clojure
:death
(let [[tw th] (term/size)
      runes (render/death-runes tw th)
      replay-code (replay/encode-run world)   ;; throws if action-log has maps
      …]
```

`encode-run`'s `(every? keyword? actions)` guard from #66 was correct *before* this PR — the v1 wire format can't carry map actions, so refusing was the safest exit. After #67, any run that touches dispatch with `:equip`/`:unequip`/`:drop`/`:enchant`/`:use-item` (i.e. essentially every real run that engages with the inventory) ends with a map action in `:action-log`. So a player who dies after touching the menu now hits the death screen and throws inside `encode-run` instead of seeing the death animation.

Two paths to fix:
- **Implement a map-capable v2 of the wire format here**, so encode-run can serialize `{:type :equip :id …}` etc. Cleanest — closes the contract end-to-end across both dispatch/replay and ?replay= in one PR.
- **Make the death-screen path tolerate "no shareable code for this run":** wrap `encode-run` in `try/catch` or guard on `(every? keyword? actions)` before calling it, and render a "seed-only — replay code not available for this run" line in place of the wrapped code. Smaller change; defers the wire-format extension to a follow-up.

Either is fine, but one of them needs to land in #67 — otherwise a regression test driving dispatch through item-use and then death would crash today.

**Finding #2 is a verified screen-state divergence, not an open question.** I asked whether `:down` would route to player-move or submenu-nav after an enchant-cancel — looked at it more carefully and the divergence is real because the cancel itself is outside dispatch:

- `world.lg:1116` — scroll use sets `:screen :quick-menu :menu-action :enchant :enchant-scroll-id` (via the dispatched `:use-item`, so the world advance + the log entry are both there)
- `main.lg:240-241` — the ESC cancel is `(dissoc world :screen :menu-action :menu-sel)`, applied as a bare world mutation by `quick-menu-input` — not via dispatch, not logged, no seed advance

So a recorded sequence of `[..., {:type :use-item :id <scroll>}, <next game action>, ...]` plays back as:

- Live: use-item → submenu → ESC (off-log) → game-screen → dispatch next game action from `:screen :game`
- Replay: use-item → submenu → (no cancel in log) → dispatch next game action from `:screen :quick-menu :menu-action :enchant :enchant-scroll-id <scroll>`

`menu_dispatch_test.lg:20-23`'s round-trip compares only `:entities`, so this divergence is invisible to the test. A test that drives `[use-scroll, ESC-cancel, move]` and compares full world (`:screen`, `:menu-action`, `:enchant-scroll-id`) would pin it.

Two ways to close it:
- **Route the cancel through dispatch** as a logged action (e.g. `{:type :cancel-menu}`) that the replay can apply. Most consistent with the widen-the-contract framing.
- **Narrow the claim:** acknowledge in `replay.lg`'s docstring that enchant-scroll cancel paths still diverge in `:screen` state, even though `:entities` is preserved. Smaller scope; pairs naturally with the inscribe/hazard follow-up.

Sorry for the soft framing in the first pass — should've traced through to the actual call sites before posting. The rest of the review (failed-drop `:inv-sel`, no-op seed-advance, multi-id dispatchable action note, test coverage suggestions) stands as-is.

nnunley:
Added a fix commit: **`fix(death): degrade replay code to seed-only for non-keyword runs`**.

This widen logs item actions as parameterized map actions (`{:type :use-item :id …}`). The death screen called `replay/encode-run`, whose keyword-only guard *throws* on such a log — so **dying after using any item crashed the game** with a FATAL ERROR. This is a direct consequence of the widen, so the guard belongs here.

Fix: new `replay/encode-run-safe` returns `nil` (instead of throwing) for non-keyword logs; the death screen uses it and already skips a `nil` code, showing the **seed line alone** — which still reproduces the opening state, per the documented replay-scope caveat. Covered by two new tests (`encode-run-safe` encodes keyword runs; returns `nil` for map-action logs). 251 tests / 0 fail; reproduced the original crash path headlessly and confirmed it now renders cleanly.

mparrett (commented):
**#1 fix looks right.** `encode-run-safe` is the cleaner of the two paths — small surface area, the guard at `replay.lg:212` is tight, and the death screen already had `(when (and replay-code (> (count replay-code) 0)) …)` around the wrapped block so nil degrades naturally to "seed line only." Players still get the seed-reproducible opening state for those runs.

Two small notes on the new tests:
- `encode-run-safe-returns-nil-for-map-actions` uses a synthetic `{:seed-input … :action-log [...]}` — proves the guard, but not the integration path. The "real" end-to-end (dispatch through an item action, then check that `encode-run-safe` of the resulting world returns nil) is also worth a one-liner; you've already got `dispatch/replay 79 30 12345 [:wait]` style helpers, so it's cheap to add.
- The death-screen rendering of "seed line, no wrapped code" isn't asserted anywhere. Probably fine — it's a render path, not a determinism path, and the Layer-1 smoke would catch a render-throws regression — but if you want to pin the UX explicitly, a snapshot of the seed-only death screen alongside the keyword-run one would do it.

**Still open from the prior review:**
- **#2 — enchant-cancel screen-state divergence** (`main.lg:240-241` cancel outside dispatch). Real correctness gap; in the same class as the inscribe/hazard-confirm follow-up.
- **#3 — failed-drop `:inv-sel` decrement** (always-decrement at `main.lg:75-78`). Cosmetic.

Both are reasonable follow-up material if you'd rather keep this PR scoped to the death-screen fix. Defer to you / @nooga on whether they block merge.

LGTM modulo those two items.

nooga:
I'm happy with this. I think we can keep the steady pace and address those two outstanding items later.

nooga (approved):

nnunley:
Follow-up commit: **`replay: carry parameterized map actions in the wire format`** — makes item runs fully replayable again (closes the gap the previous commit only guarded against).

Per your call, **no version bump**: a dictionary entry's length byte uses `len == 0` as a sentinel for a parameterized map action (`{:type :id (:scroll-id)}`), since keyword names are never empty. So **keyword-only codes stay byte-identical** — the existing e2e `REPLAY_CODE` still decodes unchanged (verified) — while item use/equip/unequip/drop/enchant now round-trip exactly. `encode-run` drops its keyword-only guard; `encode-run-safe` becomes a non-throwing wrapper (defense-in-depth for the death screen). The death screen now shows a real replay string for item runs again.

Tests: encode/decode round-trip of a mixed log (incl. enchant's `:scroll-id`), `encode-run` on an item run, and the back-compat decode. 253 tests / 0 fail.

PR #68, opened 2026-06-06
Responsive UI event loop: poll-based run-loop for animated screens (#56)

nnunley:
Implements #56 (design: #55 / `docs/responsive-ui-event-loop-design.md`). Stacks on #66.

## Problem
The determinism refactor made the game loop purely input-driven via blocking `read-key`. That froze idle animation (title/death particles), and in wasm it's worse: `read-key` is backed by `Atomics.wait` on a SharedArrayBuffer and Go-wasm is single-threaded, so a blocking read freezes the **entire** worker. mparrett's #61 patched title/death with a one-off `key-pending?` poll-peek; this generalizes that into a small, composable event-loop module so the same loop drives every animated screen, native and wasm, without leaking cosmetic ticks into the deterministic `(seed, action-log)` replay stream.

## What's here (4 commits)
- **PR1 — `xsofy.ui/run-loop` + sources.** A poll-based event loop: a screen lists non-blocking **sources** (`input-source` drains queued keys → `[:action …]`; `clock-source` emits `[:tick frame]`; `replay-source`; `poll-source`) and a `step` scan over `{:world :ui}` that returns `(reduced …)` to exit. Cosmetic `[:tick]` advances `:ui` only — never the seed/log. Includes `ui_test.lg` with a determinism property test (interleaving ticks doesn't change `:world` / `(seed, action-log)`).
- **PR2 — title + death screens** on `run-loop`; deletes the one-off `make-key-poller`/`drive-until-key`.
- **PR3a — core play loop** on `run-loop` (input + clock-driven auto-actions / ambient vfx), folding the between-turn vfx path into the unified `:ui`.
- **PR4 — docs.** `docs/terminal.md` "Responsive UIs" section: the poll-on-tick source contract, why blocking `read-key` freezes wasm, and the determinism caveat.

## Notes
- No let-go `poll!`/`alts!` exists (v1.9.0), so async/channel sources are documented as a seam, not built (FR filed upstream: nooga/let-go). `run-loop` paces via the wasm-aware `pause!` (`<!! (timeout (max ms 30))`).
- **Modal screens** (inventory/quick-menu/messages/help/inscribe/confirm) still use blocking `read-key` — that's the deferred PR3b (modal screens → nested run-loops), a follow-up.

## Verification
`make test` green (259 tests / 0 fail); `make smoke-lg` (frozen-title guard); `make e2e` (title animates in headless wasm — the #61 regression gate).

### DIFF
--- a/docs/terminal.md
+++ b/docs/terminal.md
@@ -208,11 +208,80 @@ For better performance, batch a full row into a single string:
       (recur (update-state state key)))))
 ```
 
+## Responsive UIs: multiplexing input and a cosmetic clock
+
+The blocking game loop above is fine for a turn-based screen where **nothing
+moves while you wait for input**. The moment a screen needs to *animate while
+idle* — a pulsing title, drifting particles, ambient effects — a blocking
+`read-key` falls apart:
+
+- **Native:** the loop is stuck in `read-key` until a key arrives, so frames
+  never advance and the animation freezes.
+- **WASM (browser):** far worse. `read-key` is backed by `Atomics.wait` on a
+  SharedArrayBuffer, and Go-wasm is single-threaded/cooperative, so a blocking
+  read freezes the **entire** worker — including any timer goroutine driving
+  animation. The screen never paints past frame 0.
+
+The fix is a **poll-based source contract**: never block while idle. Peek input
+with the non-blocking `term/key-pending?` and only call `read-key` once a key is
+actually queued; meanwhile a cosmetic clock advances the animation. The same
+loop then runs identically native and in wasm.
+
+xsofy implements this as `xsofy.ui/run-loop` (see
+`docs/responsive-ui-event-loop-design.md`). A screen describes **which sources
+it listens to** and **how state folds over events**; the loop drives render:
+
+```clojure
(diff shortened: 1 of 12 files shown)


mparrett:
One thing worth tightening before landing: `input-source` now maps a `nil` read to `[:action :eof]`, but the current dispatcher only treats `nil`, `:unknown`, and `:resize` as ignored inputs. Once this branch is combined with `browser-gate`, an EOF event can flow through the normal deterministic dispatch path, advance the seed, and append `:eof` to `:action-log` even though the world/entities do not change.

I verified this locally against the stack tip with a direct dispatch probe: `dispatch :eof` changed `:seed` and produced `[:eof]` in the log while leaving entities equal. That seems like the same class of non-action as `:resize`, so I think either `:eof` should be added to the ignored action set, or the play loop should turn it into a quit/exit before it reaches deterministic dispatch.

A small regression test next to the existing ignored-action tests would pin this nicely: `:eof` should leave `:seed`, `:action-log`, and `:screen` unchanged unless the intended behavior is an explicit quit path.

nooga (approved):

PR #69, opened 2026-06-06
Land the determinism + replay + repro/e2e stack onto main (#63–#66, then #67/#68)

nnunley:
## Why this PR
The stacked PRs #63–#66 were each merged into the **next branch in the stack** rather than cascading to `main` (a side effect of using branch bases + a squash-merge on #66). As a result all that approved work accumulated on `browser-gate` but `main` is still at #62. This PR brings the whole approved stack onto `main` in one step.

## What lands now (already approved + merged onto `browser-gate`)
- **#63** — repro test infra: Layer-1 runtime smoke + `?seed=` browser bridge
- **#64** — swap to native xxh3 (value-preserving)
- **#65** — headless-browser e2e gate (boot + seeded regression)
- **#66** — seed display + shareable replay codes (`?replay=`)

## What joins automatically
**#67** (widen replay contract: menu item-actions through dispatch + death-fix + map-action codec) and **#68** (#56 responsive UI event loop) both target `browser-gate` and are now mergeable. Once they merge into `browser-gate`, their commits appear in this PR — so merging this last brings #63–#68 to `main` together.

## Verification
The tip of this branch builds and passes `make test` (green), `make smoke-lg`, and `make e2e` (4/4) — these were the gates on each constituent PR.

## Merge order
1. Merge **#67** and **#68** into `browser-gate`.
2. Merge **this PR** (`browser-gate → main`).


mparrett:
I found one behavioral regression in the quick-menu item flow while reviewing the stack: using an enchant scroll from the `:use` quick-menu closes the menu instead of transitioning into the enchant target selection.

The issue looks like the cleanup after `execute-menu-action`: it preserves `result` only when `(:screen result)` differs from `old-screen`. `world/use-item` for `:enchant` intentionally returns `:screen :quick-menu` with `:menu-action :enchant` and `:enchant-scroll-id`, but because the screen is still `:quick-menu`, the cleanup strips `:screen`, `:menu-action`, and `:enchant-scroll-id` and returns to the game.

I reproduced this locally by giving the player a `:scroll-enchant`, opening `{:screen :quick-menu :menu-action :use}`, and selecting the scroll letter. The result had `:screen nil`, `:menu-action nil`, `:enchant-scroll-id nil`, and `:action-log [{:type :use-item, ...}]`.

Suggested fix: preserve the result when it still has a screen but changed menu mode/state, not only when the screen keyword changes. A regression test for "use enchant scroll from quick menu opens `:menu-action :enchant` with `:enchant-scroll-id`" would cover the path missing from the current green suite.

nooga (approved):

PR #70, opened 2026-06-06
Extract interaction loop into xsofy.menu + xsofy.game; thin main (#56 PR3b)

nnunley:
## What

Splits the monolithic `main.lg` interaction loop into focused, testable modules and migrates every remaining modal screen off blocking `read-key` onto the poll-based `run-loop` (the wasm-safe event loop from #56).

- **`xsofy/menu.lg`** — inventory/quick-menu/inscribe handlers, extracted from `main`.
- **`xsofy/game.lg`** — the screen router (`run`), `play-replay`, modal-loop helpers (`prompt-loop`, `modal-input-loop`), and game setup.
- **`xsofy/play.lg`** — the poll-based turn loop.
- **`main.lg`** — now a thin ~44-line bootstrap: terminal lifecycle + `-main`.

## Why

The interaction loop was bound to the `(-main)` boot hook, so it couldn't be unit-tested. Extracting it into modules makes the router and each modal handler testable in isolation (see the new `game_test`, `menu_test`, `play_test`, `ui_test`), and completes #56 PR3b — **no screen blocks on `read-key` anymore**, so single-threaded wasm never freezes on `Atomics.wait`.

## Also in this branch
- `play-replay` driven by the poll-based loop (a keypress during replay can no longer wedge wasm).
- Game layout anchored upper-left (was centered).
- Replay codec carries parameterized map actions; death screen degrades to seed-only via `encode-run-safe` instead of crashing.
- `make serve` target + `Cache-Control: no-store` so rebuilds aren't masked by a cached bundle.
- `docs/terminal.md` Responsive UIs section.

## Verification
- `make test` — **273 tests, 0 fail**.

## Stacking note
Based on `browser-gate`. Shares the #56 event-loop surface with #68 — once #68 and #69 land, this retargets cleanly onto `main` (overlap reconciles, leaving just the module split + PR3b). Rebase onto fresh `main` after the stack lands.


mparrett:
I did a pass over this extraction branch too. The broad shape looks good to me: the split into `xsofy.game`, `xsofy.menu`, and `xsofy.play` makes the loop pieces much easier to test, and `make test` passed locally at the branch tip (`273` tests, `0` failures).

Two practical notes before this lands:

- GitHub currently reports this PR as dirty against `browser-gate`, so it will need the rebase/retarget described in the stacking note after #68/#69 settle.
- This branch carries the same quick-menu enchant-scroll cleanup behavior I noted on #69, now in `xsofy/menu.lg`: using an enchant scroll from the use quick-menu closes the menu instead of preserving `:menu-action :enchant` / `:enchant-scroll-id`. If that gets fixed lower in the stack, this branch should pick it up cleanly during the rebase; otherwise it is worth adding a targeted menu test here as part of the extraction coverage.

nooga (approved):

nooga (approved):

PR #76, opened 2026-06-09
feat(input): add Ctrl-C as a confirm-quit alias for Esc

mparrett:
## What

Binds **Ctrl-C** to the existing `:quit` action, so it opens the same `y/n`
`:confirm-quit` prompt that `Esc` does. `Esc` is unchanged. Ctrl-C does **not**
quit immediately.

*(This started life as a bare-`q` PR, then an immediate-Ctrl-C experiment. Both
are dropped; see the thread. This is the safe version.)*

## Why this shape

- Quitting a permadeath roguelike should stay deliberate. `Esc` + `y/n` is already
  on the light end for the genre (NetHack confirms, Brogue menus, DCSS makes you
  type), so the default needs no loosening. This only adds a second, familiar key
  into the same safe path.
- `Ctrl-C → "really quit?"` is the NetHack convention, so it matches what genre
  players already reach for.
- Deliberately **not** an immediate/unconfirmed exit. That would reintroduce the
  accidental-run-loss problem that retired bare `q` after the HN launch, just
  behind a different key.

## How it works (the non-obvious bit)

`term/raw-mode!` (golang.org/x/term `MakeRaw`) disables `ISIG`, so Ctrl-C is not a
`SIGINT`. It arrives as the byte `0x03` from `term/read-key`, where it was
previously parsed to `:unknown` and dropped. Binding it is just adding the key to
the map.

The change is one line: `(str (char 3))` joins the `:quit` binding's `:keys`, and
`"Ctrl-C"` joins `:help-keys`. `(str (char 3))` keeps the raw control byte out of
the source. The keybindings screen (`?`) derives labels from `:help-keys`, so it
now shows `Esc, Ctrl-C` for free.

## Tests

`input_test.lg` asserts both `Esc` (`0x1b`) and `Ctrl-C` (`0x03`) resolve to
`:quit`, and that the help label reads `Esc, Ctrl-C`.

## Deferred: configurable keybinds

Per @nnunley, user-remappable keys are the real long-term answer (and @nooga is
right that the binding *data model* is already a plain map). What's missing is a
persistence layer (file native, localStorage in WASM) and a rebind UI. Tracked
separately rather than blocking this.

### DIFF
--- a/xsofy/input.lg
+++ b/xsofy/input.lg
@@ -28,7 +28,7 @@
    {:action :rune-codex  :category :information :label "Open rune codex"  :keys ["r"] :help-keys ["r"]}
    {:action :messages    :category :information :label "Open message log" :keys ["m"] :help-keys ["m"]}
    {:action :help        :category :information :label "Open keybindings" :keys ["?"] :help-keys ["?"]}
-   {:action :quit        :category :system     :label "Quit"              :keys ["\u001b"] :help-keys ["Esc"]}])
+   {:action :quit        :category :system     :label "Quit"              :keys ["\u001b" (str (char 3))] :help-keys ["Esc" "Ctrl-C"]}])
 
 (defn key-matches? [k binding]
   (some (fn [key] (= k key)) (:keys binding)))
--- a/xsofy/test/input_test.lg
+++ b/xsofy/test/input_test.lg
@@ -1,0 +1,15 @@
+(ns xsofy.test.input-test
+  "Tests for xsofy.input — key→action resolution and help-key labels."
+  (:require [test :refer [deftest is testing]]
+            [xsofy.input :as input]))
+
+(deftest quit-binds-esc-and-ctrl-c
+  (testing "both Esc (0x1b) and Ctrl-C (0x03) resolve to :quit"
+    (is (= :quit (input/parse-key (str (char 27)))))
+    (is (= :quit (input/parse-key (str (char 3))))))
+  (testing "the keybindings screen advertises both, Esc first"
+    (is (= "Esc, Ctrl-C" (input/action-key-labels :quit)))))
+
+(deftest unbound-keys-resolve-to-unknown
+  (testing "a nearby unbound key is :unknown, not a stray action"
+    (is (= :unknown (input/parse-key "Q")))))
--- a/xsofy/test/run.lg
+++ b/xsofy/test/run.lg
@@ -17,6 +17,7 @@
             [xsofy.test.replay-test]
             [xsofy.test.ui-test]
             [xsofy.test.play-test]
-            [xsofy.test.menu-dispatch-test]))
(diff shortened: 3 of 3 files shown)


nooga:
Not sure about this one. Initially bare `q` was used instead of `esc`. I've changed it after the game launched on HN and people complained about accidentally ending their runs trying to *q*uaff a potion. We have the y/n confirmation now so it's less of a problem. 

I think we can merge it and see how it goes - worst comes to worst can always revert.

nnunley:
So... As the voice of feature creep 🙃 , perhaps the compromise is user key bindings? A menu that lets you rebind the various keys could give you what you want, with some classic bindings matching brogue, nethack, etc, might give better affordances to people.

nooga:
@nnunley IIRC the keybinds are already stored as a map thanks to our beautiful programming language... so this would be mostly a UI bikeshed, right?

mparrett:
@nooga Good context on "why not `q`" -- that makes sense. And appreciate @nnunley 's bikeshedding awareness... probably worth disclosing my motivation, probably not articulated in the bot's write-up:

I am currently someone frequently entering and exiting the game (for testing) -- and my sensibility is that `esc`+`y` / `esc`+`n` is a bit clunky for my needs. Totally appropriate as a default exit key; I'm not trying to `q`uaff potions at all right now. I could live with `mod` + `q` or `mod` + `x` (could be shift?), but I might be in the minority, and maybe worth deferring until eventual configurable keybind design. An editable config might get the job done for ~picky~ particular people like myself.

In the meantime, how do we feel about Ctrl-C? :)

mparrett:
Followed up on my own Ctrl-C question. I prototyped it as an *immediate* exit and then talked myself out of it: an unconfirmed quit in a permadeath game just re-opens the accidental-loss door you closed after HN, only behind a different key. Not worth it.

Where I landed: keep `Esc` + `y/n` as the default. It's honestly already on the light end for the genre (NetHack confirms, Brogue/DCSS make you work harder), so it doesn't need loosening. The one safe addition is **Ctrl-C as an alias into that same confirm prompt**, NetHack-style "Ctrl-C → really quit?". Costs a line: Ctrl-C already arrives as a `0x03` byte under raw mode, it was just being dropped as `:unknown`. No new exit semantics, just a second familiar key.

That doesn't fix my original gripe (fast exit while testing), but that's a *me* problem, not a game-default problem. I'll solve it out of band rather than push my dev loop into everyone's keymap.

Reworking this PR to the Ctrl-C → confirm version. And +1 to @nnunley on configurable keybinds as the real long-term answer for picky people like me; parking that as a separate follow-up rather than scope-exploding this one.

nnunley:
LGTM, though I'm realizing that exiting on non-death should probably dump the replay key. At least until we have saved game state.

nooga (approved):

PR #77, opened 2026-06-13
Publish Homebrew formula to shared tap

nooga:
## Summary
- add tag release workflow that updates Formula/xsofy.rb in nooga/homebrew-tap
- install xsofy sources with a launcher that runs lg main.lg
- update install docs to use nooga/tap

## Checks
- ruby YAML parse for release workflow
- git diff --check

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,77 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+permissions:
+  contents: read
+
+jobs:
+  homebrew:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Update Homebrew formula
+        env:
+          GH_TOKEN: ${{ secrets.GORELEASER_PAT }}
+          TAG: ${{ github.ref_name }}
+        run: |
+          set -euo pipefail
+
+          version="${TAG#v}"
+          branch="xsofy-${version}"
+          url="https://github.com/nooga/xsofy/archive/refs/tags/${TAG}.tar.gz"
+          sha256="$(curl -fsSL "$url" | sha256sum | awk '{print $1}')"
+
+          git clone "https://x-access-token:${GH_TOKEN}@github.com/nooga/homebrew-tap.git" tap
+          cd tap
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git checkout -B "$branch"
+          mkdir -p Formula
+
+          cat > Formula/xsofy.rb <<RUBY
+          class Xsofy < Formula
+            desc "Roguelike written in let-go where the magic system is Lisp"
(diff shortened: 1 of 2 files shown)


mparrett:
Drive-by, same as on let-go #218 — this looks good. The source-install shape is right: `libexec` holds `main.lg` and the `xsofy/` tree, the `bin/xsofy` launcher execs `lg main.lg "$@"`, and `let-go` comes in as a dependency. The idempotent re-run guard (`git diff --cached --quiet`) and keeping the cross-repo write on `GORELEASER_PAT` while the workflow token stays `contents: read` are both clean.

`depends_on "let-go"` is a bare formula name, so it resolves only once `nooga/tap` is tapped. `brew install nooga/tap/xsofy` auto-taps it, so it should work, but cross-tap deps are safer fully qualified as `depends_on "nooga/tap/let-go"`. A clean-machine `brew install nooga/tap/xsofy` would confirm the dep resolves.

Two small notes: the `test do` block asserts `libexec/"xsofy/title.lg"`, so that path needs to be in the release tarball; and as on #218, `GORELEASER_PAT` needs write plus `--admin` merge rights on `nooga/homebrew-tap`.

nooga:
Updated the generated formula to use the fully qualified dependency: `depends_on "nooga/tap/let-go"`.

The test path note is covered: `xsofy/title.lg` is tracked and included in source archives. The tap permissions note should also be covered for the configured token path; the current GitHub identity has admin/write on `nooga/homebrew-tap`, and `GORELEASER_PAT` is set on this repo.

mparrett (approved):
Thanks for addressing the feedback. Overall this looks good. Noting one thing you are probably already tracking:

The two PRs share one operational dependency, which needs to be sequenced like this:

1. Cut a let-go release (v1.10.1, or re-tag) and its workflow seeds Formula/let-go.rb. This also un-breaks both READMEs.
2. Then the first xsofy v* tag via #77's workflow writes Formula/xsofy.rb, and its `depends_on "nooga/tap/let-go"` resolves.

If xsofy gets tagged before a post-#218 let-go release, `brew install nooga/tap/xsofy` fails on the dependency.

mparrett:
@nooga One other thing worth discussing soon (not a blocker here):

**TL;DR**- current minimum version gating in CI doesn't cover the shipped version, which may be newer (and potentially incompatible). Worth discussing our desired contract here and making it explicit.

The formula as written here depends on `nooga/tap/let-go` unpinned, and we currently gate/pin on PR/manual dispatch, not at release time. This means the tagged source that becomes a brew formula is never re-tested at publish.

Observations worth mentioning:
- CI pulls let-go from the Go module proxy (@v1.9.0 / @latest). The brew formula installs let-go from the tap (goreleaser-built binary of let-go's latest release). Different acquisition path.
- The brew dep floats to latest let-go. CI's required leg is v1.9.0 (the floor), which protects against let-go being too old. Nothing blocking protects against it being too _new_ , which is the direction the unpinned brew dep actually moves.

cc: @nnunley

PR #78, opened 2026-06-17
build(wasm): client-owned shell via lg -w -w-shell none

mparrett:
## What

Build the WASM bundle glue-only (`lg -w -w-shell none`) and inject xsofy's **own** shell, instead of patching let-go's generated artifact. The shell binds to the runtime through the public `window.LetGoHost` surface (`onReady` / `onOutput` / `sendInput` / `setSize`) — no reaching into let-go's internals. xsofy owns the terminal, theme, and font, ~including~ (deferred; see #79) a bundled monospace face for the Runic-block glyphs (U+16A0–16FF) that previously had no clean home.

- `tools/xsofy-shell.html` — xsofy's xterm shell + `@font-face` hook + `onReady` binding
- `tools/inject_shell.lg` — append the shell before `</body>` (idempotent, fail-closed, same shape as the `patch_wasm_*` tools)
- `Makefile` — `-w-shell none` + the inject step

## Why this is the win

Custom presentation was the motivating case for shell extraction: anything beyond let-go's default shell — embedding fonts, wiring an audio hook, rendering graphics next to the terminal — meant patching the generated artifact. With the client owning the shell, that's all just xsofy's own code now: no patch, and nothing let-go needs to know about.

Validated interactively on **iOS Safari** over a header-capable HTTPS serve: the game renders and takes input in xsofy's own shell, bound only through `LetGoHost.onReady`.

## Scope — honest about what this does NOT remove

This replaces **shell/presentation** patching. Two patches stay, because they touch concerns the shell contract doesn't cover:

- `patch_wasm_coi.lg` — COI bounded-retry for header-less hosts (GitHub Pages). Touches the COI lifecycle; core owns service-worker registration.
- `patch_wasm_seed.lg` — `?seed=`/`?replay=` → worker `go.env`. xsofy reads the seed via `os/getenv` in the **worker** VM; passing env there isn't in the `LetGoHost` surface.

Both now patch let-go's *core glue* (stable markers) rather than its shell. They motivate follow-up seams — an env/config surface and a COI option — tracked on the let-go side, not blocking this.

## Companion

let-go side: [nooga/let-go#245](https://github.com/nooga/let-go/pull/245) (merged) — `feat(wasm): client-owned shell — split lg-host.js, add -w-shell none`. This PR is the client that proves that contract.

### DIFF
--- a/tools/xsofy-shell.html
+++ b/tools/xsofy-shell.html
@@ -2,26 +2,16 @@
      tools/inject_shell.lg. Binds to the let-go runtime only through the public
      window.LetGoHost surface (onReady / onOutput / sendInput / setSize) — no
      patching of let-go's generated glue. xsofy owns the terminal, theme, and
-     font here; in particular the Runic-block glyphs (U+16A0-16FF) get a bundled
-     monospace face instead of the platform's proportional fallback. -->
+     font. (Runic-block glyphs (U+16A0-16FF) fall back to the platform face here;
+     a dedicated rune font is a follow-up PR stacked on this one.) -->
 <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@xterm/xterm@5.5.0/css/xterm.min.css">
 <script src="https://cdn.jsdelivr.net/npm/@xterm/xterm@5.5.0/lib/xterm.min.js"></script>
 <script src="https://cdn.jsdelivr.net/npm/@xterm/addon-fit@0.10.0/lib/addon-fit.min.js"></script>
-<style>
-  /* The held rune font (Fairfax HD, OFL). Ship fonts/FairfaxHD.woff2 alongside
-     the bundle; under COEP require-corp it must be same-origin (it is). Falls
-     back to a system monospace until the woff2 is added. */
-  @font-face {
-    font-family: "Fairfax HD";
-    src: url("fonts/FairfaxHD.woff2") format("woff2");
-    font-display: swap;
-  }
-</style>
 <script>
 (function () {
   const termEl = document.getElementById('terminal');
   const term = new Terminal({
-    fontFamily: '"Fairfax HD", "DejaVu Sans Mono", "Menlo", monospace',
+    fontFamily: '"DejaVu Sans Mono", "Menlo", monospace',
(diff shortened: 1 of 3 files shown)


tools/xsofy-shell.html
```diff
@@ -0,0 +1,46 @@
...
+<script src="https://cdn.jsdelivr.net/npm/@xterm/addon-fit@0.10.0/lib/addon-fit.min.js"></script>
+<style>
+  /* The held rune font (Fairfax HD, OFL). Ship fonts/FairfaxHD.woff2 alongside
+     the bundle; under COEP require-corp it must be same-origin (it is). Falls
+     back to a system monospace until the woff2 is added. */
+  @font-face {
+    font-family: "Fairfax HD";
+    src: url("fonts/FairfaxHD.woff2") format("woff2");
+    font-display: swap;
+  }
```
mparrett: This is something I was playing with, kept it here for a demo, but likely wouldn't make it into the non-draft version of this PR.

nnunley: I'd suggest using a web font CDN -- it looks like FairFaxHD is available:

&lt;link href="https://cdnfonts.com" rel="stylesheet"&gt;

And then you can drop the src.

nooga (approved):

nooga:
Exciting! 🤩 

Let's get this merged @mparrett !

tools/xsofy-shell.html
mparrett: Good call. Thanks!

mparrett:
> Exciting! 🤩
> 
> Let's get this merged @mparrett !

Thanks for the support, glad the overall shape fits! I'm going to take a closer look at the wiring and smoke test a bit more. There are a few minor things up in the air I'm working to resolve before we land this. It's coming together!

tools/xsofy-shell.html
mparrett: Thanks for this. We pulled the rune font out of this PR into a follow-up so #78 stays focused on the shell — it's #79. The delivery question (CDN vs self-hosted subset) is being worked out there.

nnunley (approved):

PR #79, opened 2026-06-19
feat(shell): Fairfax HD as the single terminal font (fixes rune cell-advance drift)

mparrett:
Stacks on #78 (client-owned shell), now merged.

## What this does

Renders the xsofy terminal in a single bundled face — **Fairfax HD** — so every glyph (ASCII, box-drawing, runes, symbols) shares one cell advance.

## Why single-face, not a scoped rune font

The motivation is unchanged: the Runic-block glyphs the game draws have no usable platform fallback — proportional on macOS (overflow the cell) or absent elsewhere (tofu) — so the game needs a monospace face with Runic coverage bundled in.

The first approach here was to bundle Fairfax HD scoped to the Runic block with `unicode-range`, leaving ASCII/UI on the fallback mono. A self-review caught the flaw: **Fairfax HD is a *halfwidth* monospace** — narrower cells than the fallback mono. Both are monospace, but they have different advances. With `unicode-range`, xterm measures the cell from the fallback mono (~9.63px) while runes render from Fairfax (~8px), so every rune is ~1.6px narrow and the columns after it drift left. Worse, the fallback mono is platform-dependent, so the mismatch varies by device. The `document.fonts.ready` refit doesn't help — it re-measures the unchanged ASCII cell.

The drift is structural to mixing two monospace fonts of different widths, so the fix is to not mix: **one face for the whole terminal.** Fairfax HD is built for this: a halfwidth monospace terminal font with broad coverage (Latin, pseudographics/box-drawing, runes, symbols), so xterm measures and renders everything from its one advance, and the grid can't drift.

## Packaging

Still fully self-contained: the subset is inlined as a base64 WOFF2 in the shell's `@font-face`, baked into the single `index.html` at build time — no external request, same-origin under the bundle's `require-corp`.

The base64 is **generated**, not hand-maintained. `tools/gen-terminal-font.py` (`make rune-font`) subsets Fairfax HD to printable ASCII + the full Runic block + every non-ASCII glyph `xsofy/*.lg` renders, then rewrites the `@font-face` in place. So the subset can't fall out of sync as the game's glyph set grows — re-run it after adding a glyph. The source TTF (3.2 MB) isn't vendored; the script downloads it from a pinned, checksum-verified `kreativekorp/open-relay` commit and caches it (gitignored). Current subset: 226 glyphs, ~7 KB WOFF2 / ~9.5 KB base64 — a rounding error against the WASM payload.

The tradeoff is that the whole terminal now adopts Fairfax HD's halfwidth, slightly pixel-ish look rather than just the runes. That's intended.

## Validation

- Built the bundle and rendered it in xterm: ASCII, runes, box-drawing, and symbols all render from the one face; the game boots and draws runes correctly.
- Measured advances in the live terminal: `M`, `─`, `ᚠ`, `①` are all equal — uniform cell, no drift.
- `tools/gen-terminal-font.py` asserts every glyph the game renders survives the subset (it exits non-zero if Fairfax lacks one), so coverage is checked against `xsofy/*.lg`, not assumed.


mparrett:
The font-delivery method is still open, worth settling before this leaves draft.

The CDN `<link>` works, but the platform has moved against shared font CDNs: cache partitioning (Chrome 86+, Safari, Firefox) removed the cross-site cache benefit, third-party CDNs add a privacy/IP-leak consideration, and cdnfonts serves WOFF1 rather than WOFF2. For this case a self-hosted runic-subset WOFF2 is also simpler under COEP — same-origin needs no `crossorigin` — and it can be scoped to `unicode-range: U+16A0–16FF` so only runes use the face (the linked CDN stylesheet is full-coverage and can't be scoped).

Leaning toward self-hosting a subsetted WOFF2; keeping the CDN form here until we land that.

mparrett:
Added a note about motivation to the PR description.

mparrett:
**Delivery method resolved** (the open question from the original description).

Two decisions, both backed by an in-browser A/B in real xterm 5.5.0:

- **CDN → inlined self-host.** Cache partitioning killed the shared-CDN benefit, and inline base64 is same-origin so it sidesteps the COEP `crossorigin` requirement entirely and removes the third-party dependency. ~2.4 KB inlined is negligible next to the WASM payload.
- **Fairfax HD over Noto Sans Runic.** Both render correctly in xterm (the fixed-cell DOM renderer holds the advance, so even Noto doesn't overflow), but Fairfax reads clearer — better-proportioned and laid out, and noticeably less jumbled than the proportional system fallback. Noto was viable but second.

The comparison harness (Noto vs Fairfax vs system fallback, multiple sizes + a compact 5×5 rune grid, all in xterm) lives in our workspace if useful.

Subset is the assigned Runic block (U+16A0–16F8, 89 glyphs) rather than xsofy's exact 25 runes, so adding a rune later won't tofu — costs ~1.4 KB over the game-exact subset.

mparrett:
*Partial Codex Findings (5.5-high):*

The inlined payload decodes cleanly as a 2,428-byte WOFF2. The native suite and WASM build are still running; I’m waiting for both before forming findings.

The WASM target built and injected the PR shell, but the two post-injection patch scripts reported getenv resolution errors with this local lg binary while still returning success. That looks like an existing tooling/runtime mismatch outside this PR’s changed files, so I’m treating it as a verification limitation and continuing with checks that isolate the new shell/font behavior.

mparrett:
**Codex Findings (5.5-high)**

- **[P1] Rune font advance does not match xterm cell width** in [tools/xsofy-shell.html](/private/tmp/xsofy-pr79/tools/xsofy-shell.html:24). Because `unicode-range` leaves ASCII on DejaVu/Menlo, xterm measures/fits cells from the fallback mono face, but runes render from Fairfax. In a browser check against the PR CSS after font load, Fairfax runes measured `8px` each while the fallback ASCII cell measured `~9.63px`. Rows containing runes will compress by `~1.63px` per rune, so following columns drift left. The `document.fonts.ready` refit does not fix this because the measured ASCII cell width is unchanged. The fix needs either matching font metrics, using one face for all terminal glyphs, or another rendering strategy that forces rune glyphs into the xterm cell advance.

**Verification**

- GitHub PR checks are green.
- `make test LG=/Users/matt/projects-new/3p/let-go/lg` passed on rerun: `Tests: 278 Pass: 2325 Fail: 0 Error: 0`.
- `make wasm` built and injected the shell, but local post-injection patch scripts hit a `getenv` resolution issue with my local `lg`; CI’s WASM smoke checks are green, so I treated that as a local verification limitation.

mparrett:
Addressed in 12b08bfb7068b8f5c7f8a0e7ff8326135612d923

mparrett:
Note: This was a preemptive spike to explore the magic/rune glyph font options. Not 100% sure we should adopt the specific hosting approach yet, especially because the TUI case boils down to "please install this font and make sure your terminal is configured to use it" — on the other hand, a self-contained and portable web experience is arguably a near necessity.

nooga (approved):

PR #80, opened 2026-06-21
fix(tools): reactivate WASM patch scripts against current let-go glue

mparrett:
Follow-up to #78. The post-injection patch scripts it shipped (`patch_wasm_coi.lg`, `patch_wasm_seed.lg`) broke as let-go's generated WASM glue evolved, so `make wasm` aborts right after `inject_shell` on a current-let-go build. This re-couples them.

## What drifted

- **Bare `getenv`.** Both scripts read `(getenv "XSOFY_WASM_INDEX")`; current let-go exposes it as `os/getenv` (`inject_shell.lg`, which still works, already uses that form). First failure point on both.
- **COI splice end-marker.** `patch_wasm_coi.lg` splices its bounded-retry block between two marker comments. The end marker `// --- Inline wasm_exec.js and WASM data ---` is now emitted as `// --- wasm_exec.js + WASM payload ---`.
- **Seed bridge anchors (×3).** `patch_wasm_seed.lg` injects the `?seed=`/`?replay=` → `go.env` bridge at three points that all moved with the new worker init protocol:
  - the init message is now `const initMsg = { ... mode: WASM_MODE ... }` then `worker.postMessage(initMsg)`, not an inline `worker.postMessage({ ... wasmGzB64 ... })`;
  - the worker's destructuring of `e.data` gained `mode` and `wasmModule`;
  - the `// Run Go WASM` comment that uniquely marked the worker's `new Go()` is gone. Re-anchored on the worker's lowercase `eval(wasmExecJS)` — the main thread uses `eval(WASM_EXEC_JS)`, so the two stay distinct.

## Deploy pin

The Pages deploy (`deploy-pages.yml`) installed `let-go@v1.9.0` and built with `let-go -w dist main.lg`, which emits the *old* glue markers — so the reactivated scripts would fail a tag deploy at the COI splice. The marker set the scripts target only exists on let-go `main` (post-#245); no release carries it (v1.10.0 predates it). So the deploy is pinned to commit `8087f99`:

- both the binary (`go install …@<sha>`) and the WASM runtime (`LETGO_SRC`) — without the latter, `lg -w` compiles the runtime from `let-go@latest` (= the old v1.10.0 glue) and mismatches the binary's worker protocol;
- `setup-go` bumped to `1.26` (let-go's `go.mod`).

A commit rather than a tag is a deliberate stopgap until let-go ships a release with the client-shell glue. The deploy still builds the default-xterm bundle — switching it to the client shell + font (#79) is a separate migration. The patch scripts apply to either bundle (they target the shared core glue), confirmed against both.

## Verification

Built against let-go tip with `make wasm`:

- Runs clean through `inject_shell` → `patch_wasm_coi` → `patch_wasm_seed`.
- COI establishes: the terminal fits the viewport (measured 34×127 on a 1280-wide window) instead of the default 80×24 it falls back to when isolation never comes up.
- `?seed=12345` reaches the game — the title screen reports `seed 12345` rather than a random one.

No behavior change to the patched output beyond matching the current glue; the injected COI and seed/replay logic is unchanged.

### DIFF
--- a/.github/workflows/deploy-pages.yml
+++ b/.github/workflows/deploy-pages.yml
@@ -23,15 +23,31 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      # Pinned to a let-go commit, not a tag: the client-shell WASM glue these
+      # patch scripts target (markers, worker init protocol) landed after #245
+      # and is not in any release yet (v1.10.0 predates it). Bump to a tagged
+      # release once one ships with it.
+      - name: Check out let-go (pinned)
+        uses: actions/checkout@v4
+        with:
+          repository: nooga/let-go
+          ref: 8087f9979f511a9dc89266c9d84a41af00d98287
+          path: let-go-src
+
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
-          go-version: '1.24'
+          go-version: '1.26'
 
       - name: Install let-go
-        run: go install github.com/nooga/let-go@v1.9.0
+        run: go install github.com/nooga/let-go@8087f9979f511a9dc89266c9d84a41af00d98287
 
+      # LETGO_SRC pins the WASM *runtime* to the same commit as the binary above;
+      # without it `lg -w` compiles the runtime from let-go@latest (the old
+      # v1.10.0 glue), which mismatches the binary's worker protocol.
       - name: Build WASM
+        env:
+          LETGO_SRC: ${{ github.workspace }}/let-go-src
         run: let-go -w dist main.lg
 
       - name: Patch COI startup guard
--- a/tools/patch_wasm_coi.lg
+++ b/tools/patch_wasm_coi.lg
@@ -3,7 +3,7 @@
             [os]))
 
 (def index-path
-  (let [path (getenv "XSOFY_WASM_INDEX")]
+  (let [path (os/getenv "XSOFY_WASM_INDEX")]
     (if (and path (> (count path) 0))
       path
       "dist/index.html")))
@@ -12,7 +12,7 @@
   "// --- Cross-origin isolation: prefer server headers, fall back to SW ---")
 
 (def wasm-marker
-  "// --- Inline wasm_exec.js and WASM data ---")
+  "// --- wasm_exec.js + WASM payload ---")
 
 (def wasm-const-marker
   "const WASM_EXEC_JS")
--- a/tools/patch_wasm_seed.lg
+++ b/tools/patch_wasm_seed.lg
@@ -6,7 +6,7 @@
             [os]))
 
 (def index-path
-  (let [path (getenv "XSOFY_WASM_INDEX")]
+  (let [path (os/getenv "XSOFY_WASM_INDEX")]
     (if (and path (> (count path) 0)) path "dist/index.html")))
 
 (defn abort! [m] (println m) (os/exit 1))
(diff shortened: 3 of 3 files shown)


mparrett:
**Codex Findings (5.5-high)**
- Blocking: [tools/patch_wasm_coi.lg](/Users/matt/projects-new/3p/worktrees/xsofy-patch-fix/tools/patch_wasm_coi.lg:15) now expects the newer `// --- wasm_exec.js + WASM payload ---` marker, but [deploy-pages.yml](/Users/matt/projects-new/3p/worktrees/xsofy-patch-fix/.github/workflows/deploy-pages.yml:31) still installs `github.com/nooga/let-go@v1.9.0` and runs `let-go -w dist main.lg` directly. That generated output has the old marker, so a tag deploy/workflow dispatch will fail at `Patch COI startup guard`. I reproduced this against an older generated artifact: `missing or repeated marker: // --- wasm_exec.js + WASM payload ---`. Fix by updating the workflow to the same current-let-go path as `make wasm`, or keep the patch backward-compatible.

**Notes**
- The PR scripts do apply cleanly against upstream current `let-go-main` shellless golden output.
- I also tested the PR seed patch against the local fork-generated bundle that includes `urlSearch`; it fails its exact anchor there. That may be outside PR `main`, but if this branch is meant to work with the local fork, preserve `urlSearch` in both init-message and worker destructuring anchors.

mparrett:
Reassessed the updated head (8d2d94d). No remaining review findings from me.

The new Pages change addresses the prior deploy blocker: the workflow now checks out let-go at 8087f9979f511a9dc89266c9d84a41af00d98287, installs the same commit, sets LETGO_SRC to that checkout, and uses Go 1.26, matching that commit's go.mod.

Verification:
- fetched the updated PR ref locally as origin/pr/80
- checked the exact pinned let-go lg-host-core.js; its COI marker, worker init message, destructuring, and eval(wasmExecJS) anchor match the patch scripts
- ran both patch scripts against that exact pinned core file; both passed
- PR checks are green: latest (upstream smoke test) and pinned (deploy floor)

The earlier urlSearch caveat only applies to the separate local fork-generated glue. Since this PR now explicitly pins the upstream let-go commit whose glue it targets, I would not block on that.

nooga (approved):

nnunley:
Probably should try using the semver code to also link to the sha OR a version > 1.10, in order to avoid other accidental deps issues?

.github/workflows/deploy-pages.yml
```diff
@@ -23,15 +23,31 @@ jobs:
...
 
+      # Pinned to a let-go commit, not a tag: the client-shell WASM glue these
+      # patch scripts target (markers, worker init protocol) landed after #245
+      # and is not in any release yet (v1.10.0 predates it). Bump to a tagged
+      # release once one ships with it.
+      - name: Check out let-go (pinned)
+        uses: actions/checkout@v4
+        with:
+          repository: nooga/let-go
+          ref: 8087f9979f511a9dc89266c9d84a41af00d98287
```
nnunley, line 34: You probably add the semver dependency to the main for xsofy for this sha OR > 1.10, provided that we think the fix is going to land in 1.11 or so.

nnunley (changes requested):

mparrett:
I asked Codex about this, and here's the lengthy response. Maybe #1 followed by #2?

Thanks, I think that makes sense. The way I understand the concern is: the workflow now pins let-go at 8087f997..., but xsofy itself does not yet say "I require this let-go build shape". So a future CI/deploy edit could accidentally run an older let-go and only fail later in the WASM patch scripts, instead of failing early with a clear version/build requirement.

A few possible ways to handle it:

1. Short-term: add a source-level `require-letgo` guard for the exact commit, and make the CI/deploy binary expose that commit. One wrinkle: plain `go install github.com/nooga/let-go@8087f997...` appears to produce `lg dev` / commit `none`, because the commit metadata is normally stamped by release ldflags. So for a SHA guard to actually enforce, the workflow probably needs to build from the checked-out `let-go-src` with ldflags setting `main.commit=8087f997...` (and maybe `main.version=dev` or similar). That would make `(require-letgo "8087f997...")` meaningful instead of warn-and-pass.

2. Longer-term: once the glue lands in a tagged let-go release, switch the guard and workflows to semver, e.g. `require-letgo ">1.10.0"` / `>=1.11.0` depending on the release. That seems like the cleaner steady state.

3. Least change: keep the workflow SHA pin and just document why. I think this is weaker, because it still leaves the app source without an early compatibility assertion.

So I think the practical route is probably #1 now, then #2 when a release exists. It also probably means updating the test workflow's required floor away from `v1.9.0`, since after adding the guard that floor should fail clearly rather than continue as the required lane.

nnunley:
Ok. I'm fine with that solution. I'm mostly worried about diverging dependences between terminal users and browser users.

mparrett:
>  I'm mostly worried about diverging dependences between terminal users and browser users.

Appreciate that you're looking out for it. I need to think more about what discipline and/or systems we need here to identify/surface/prevent such divergence.

nnunley:
I think what you have is good enough for the immediate need, however. I can give you the approval without if you want to create a follow-up design doc for how to deal with this? It might make sense to lean on the lgx binary, actually, if @nooga is up for it, and if it supports this style of generation.

mparrett:
> I think what you have is good enough for the immediate need, however. I can give you the approval without if you want to create a follow-up design doc for how to deal with this? It might make sense to lean on the lgx binary, actually, if @nooga is up for it, and if it supports this style of generation.

Yeah, I think a focused design/approach doc is warranted here. I've been poking at some ideas locally but nothing clean has emerged yet.

mparrett:
Ok, I poked at this a little bit more, and I think we can move forward. I'll accept the approval and write the follow-up design doc for the terminal-vs-browser dependency divergence, which is the bigger problem. Leaning on lgx for this style of generation is worth exploring.

A few final notes:

- Verified the pinned build locally, end to end: `lg -w` against `8087f997` plus both patch scripts (COI splice + seed bridge) succeed, so the immediate fix is good. All markers and anchors matched.

- Found out  `go install …@<sha>` binary self-reports `lg dev` and the version comes from `ldflags` the Makefile sets (`-X main.version/commit`), which module install doesn't apply. A `require-letgo` assertion against that binary warns-and-passes instead of enforcing. Making it enforce means building from the `let-go-src` checkout with those `ldflags` (the option-1 path).

nnunley (approved):

PR #85, opened 2026-06-22
fix(play): repaint on idle terminal resize, not just on keypress

mparrett:
Resizing the terminal didn't repaint the play screen — the new layout only showed up on the next keypress. The play loop is poll-based but renders only on input, so a resize sat unhandled until something woke the loop. This makes an idle resize repaint on its own.

## Why it waited for a keypress

The loop runs on a `clock-source` ticking ~30 ms, but the only render path is inside `play-advance`, and an idle `:tick` calls it solely when auto-moving (autoexplore/rest). The terminal-size poll lives in `play-advance` too. So with no input the loop spun every tick doing nothing: a `setSize` from the shell updated `term/size`, but nothing read it until a key arrived and ran `play-advance`, which saw the new size and did a full render. The death screen already polls size per `:tick`; the play loop never did.

## The fix

An idle `:tick` now compares `term/size` to the last size and calls `render-full` when it changed (or when a `:redraw?` flag is set). It runs off the dispatch path — no `dispatch`, no seed advance, no action-log entry — so replay determinism holds (`dispatch.lg` spells out why a resize must not touch the seed). `:redraw?` is there for future force-repaint triggers (theme, font) without re-deriving the size-change check.

## Validation

Counting the bytes the worker renders, so xterm reflow can't mask the result, against the deploy-pinned let-go `8087f99`: from a settled-idle state the worker emits 0 bytes; a resize with no keypress then produces a full ~77 KB frame, repeatably; `origin/main` emits 0 on the same resize. Full suite: 278 tests, no new failures (same 4 pre-existing errors).

### DIFF
--- a/xsofy/play.lg
+++ b/xsofy/play.lg
@@ -96,7 +96,13 @@
 (defn play-step
   "The play loop's scan. [:action] dispatches an explicit key (cancelling any
    auto-mode first); [:tick] advances ambient :ui and, if auto-moving, dispatches
-   one auto-step. Both route through play-advance / the deterministic dispatch."
+   one auto-step. Both route through play-advance / the deterministic dispatch.
+
+   An idle [:tick] also repaints when the terminal resized or a :redraw? request
+   is pending. The size poll otherwise lives only in play-advance, which idle
+   ticks skip — so without this a resize stays invisible until the next keypress.
+   The repaint is off the dispatch path (no seed/action-log mutation), so it's
+   replay-safe; :redraw? is the extension point for future force-repaint triggers."
   [state ev]
   (let [[tag payload] ev
         world (:world state)]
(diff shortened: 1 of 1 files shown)


mparrett:
I completed a browser smoke test and asked Codex (5.5-high) to review.

**Findings: None** The idle `:tick` resize path in `xsofy/xsofy/play.lg` repaints and updates loop metadata without mutating `:world`, seed, or `:action-log`, so the determinism contract still holds.

Verification: `make test LG=/Users/matt/projects-new/3p/joint-xsofy/let-go-main/lg` completed with exit code 0. Reported summary was `Tests: 278 Pass: 2317 Fail: 0 Error: 4`, matching the PR note about existing errors.

---

ℹ️ I'm not personally tracking the existing errors, so I need to look into that more deeply.

mparrett:
Note on the "4 pre-existing errors" in the validation section — they're a test-harness artifact, not real failures, and not from this change.

`make test` and `make smoke-lg` capture stdout through a pipe. `term/flush` calls `os.Stdout.Sync()`, which returns `EBADF` ("bad file descriptor") on a non-syncable fd like a pipe, so the render-path tests throw and the smoke driver never prints its sentinel. Run against a TTY or a regular file and it's clean: `278 tests, 2325 pass, 0 error`, smoke prints `@@SMOKE-OK@@`. Same code, different fd.

It's a let-go behavior change after v1.9.0: v1.9.0 calls `os.Stdout.Sync()` and ignores the return (green even when piped); current let-go propagates it out of `term/flush`, so it surfaces on the `8087f99` deploy pin and on `latest`. Filed upstream as nooga/let-go#308 — the fix is `term/flush` tolerating `EBADF`/`ENOTTY` from syncing a non-syncable fd.

nnunley (approved):

Issue #87, opened 2026-06-24
Proposal: Dungeon generation dev tool + optimizations


mparrett:
## Dungeon Generation Investigation

**What**: Scratchpad to look into a few dungeon-generation issues (layout / door placement, mostly). Keeping it light for now, but will expand into a proper proposal and set of PRs if the findings are meaningful.

**Motivation:**
Investigate the floating-doors-in-hallways issue and whether it's feasible now to return to the original unbounded algorithm that predated the deterministic work.

## Goals

1. Build a tool for qualitative and programmatic dungeon map generation and tuning (PR open - #88)
2. Find the root cause for the floating door issue and offer a fix
3. Investigate performance properties, edges, and offer relevant optimizations
4. Make it easier to tune dungeon generation for future game-design choices in #40.

## Pull requests

- #88 — dungeon map viewer + audit tool (merged)
- #89 — connectivity guarantee via union-find (merged)
- #91 — stairs placed on a reachable floor tile + connectivity regression test
- #92 — floating-door cleanup, floating → 0 (stacked on #91)

## Observations/Findings

Map generation can more easily be evaluated with a dedicated generation tool. This also allows qualitative inspection and extension to statistical testing. Below are specific areas under investigation.

1. Connectivity issue
2. Bounded vs unbounded quality
3. Overall efficiency
4. Overall quality

## Scratchpad

Preliminary tool outputs - map renders. Single seeds + not comparable due to generator differences. Documenting for posterity.

| Original (pre-hashing) | Bounding opt. removed |
|---|---|
| <img width="359" height="325" alt="Image" src="https://github.com/user-attachments/assets/fa13deef-4836-46bc-90eb-f58a209548fa" /> |  <img width="359" height="313" alt="Image" src="https://github.com/user-attachments/assets/8d4a78ec-74b2-44a1-b3aa-b9750a97a123" /> |

---

A tentative "fix" for the floating door issue
<img width="357" height="324" alt="Image" src="https://github.com/user-attachments/assets/11f7cd48-8faa-4122-9e40-bc4cbe365185" />

### Follow-up

- Door generation / room attachment
- Deeper analysis re: door placement

nnunley:
So, I'm noticing that most rooms in the new unbounded generation don't have doors at all.  And there's a floating door in second from the bottom hallway strip.

I think we need to do some more research on the SoTA, but the options look like:
* traditional rooms and corridors generation + A*star to dig corridors and hallways.
[...]

PR #88, opened 2026-06-24
Add a standalone dungeon map viewer + audit (tools/mapgen.lg)

mparrett:
This adds a small dev tool for looking at what the dungeon generator produces, without booting the game. Part of the https://github.com/nooga/xsofy/issues/87 exploration.

What: `lg tools/mapgen.lg` generates a level with `xsofy.terrain` and prints it to stdout as plain ASCII, or as 24-bit ANSI color (it reuses terrain's own tile tables, so the colors match the game). Above the map it prints a one-line audit:

```
seed=42  79x30  depth=3  rooms=19  doors=26  floating=17  unreach=15
```

`floating=` counts doors that aren't at a real room mouth: a door sitting mid-corridor, or in a passage that doesn't open into a room. `unreach=` counts passable tiles cut off from the rest of the level (a flood-fill from the first floor tile). Both read 0 on a clean level.

Config is via env vars so a recipe can drive it: `MAPGEN_SEED`, `MAPGEN_W`, `MAPGEN_H`, `MAPGEN_DEPTH`, `MAPGEN_COUNT` (dump N consecutive seeds), `MAPGEN_COLOR`.

```
MAPGEN_SEED=42 lg tools/mapgen.lg
MAPGEN_SEED=42 MAPGEN_DEPTH=5 MAPGEN_COLOR=0 lg tools/mapgen.lg
```

It's one self-contained file and adds no dependencies; it only uses `xsofy.terrain` (the generator and its tile tables).

Run over a range of seeds, the audit also turns up two things in the current generator: floating doors (roughly 9–24 per level) and, on a number of seeds, disconnected levels (`unreach` up to ~15) at 79×30 depth 3. Happy to follow up with fixes for both; this PR is just the tool.

## Screenshots

seed 42
image

seed 123
image

seed 42; no colors
image

### DIFF
--- a/tools/mapgen.lg
+++ b/tools/mapgen.lg
@@ -2,38 +2,28 @@
   "Standalone dungeon map generator / debug viewer.
 
    Generates a level with xsofy.terrain and dumps the grid to stdout as ASCII
-   (optionally 24-bit ANSI color), then prints a one-line audit, so you can
-   eyeball layouts and flip through seeds without booting the game. Reuses the
-   live generator and its tile tables untouched — no duplication.
+   (optionally 24-bit ANSI color), then prints a one-line audit (see
+   tools.mapaudit), so you can eyeball layouts and flip through seeds without
+   booting the game. Reuses the live generator and its tile tables untouched.
 
-   Config via env vars (so a justfile recipe can pass them):
-     MAPGEN_SEED   integer base seed         (default: random)
-     MAPGEN_W      width in tiles            (default: 79)
-     MAPGEN_H      height in tiles           (default: 30)
-     MAPGEN_DEPTH  dungeon depth             (default: 1)
-     MAPGEN_COUNT  consecutive seeds to dump (default: 1)
-     MAPGEN_COLOR  \"0\"/\"false\" disables ANSI color (default: on)
+   Config is an EDN map, given inline or as a file path argument:
+     lg tools/mapgen.lg '{:seed 42 :depth 5}'
+     lg tools/mapgen.lg my-scenario.edn
+   Keys (all optional): :seed (default random), :w 79, :h 30, :depth 1,
+   :count 1 (consecutive seeds to dump), :color true."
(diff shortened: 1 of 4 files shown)


tools/mapgen.lg
```diff
@@ -0,0 +1,180 @@
...
+        unreach (count-unreachable grid w h)]
+    (println (str "seed=" seed "  " w "x" h "  depth=" depth
+                  "  rooms=" (count rooms) "  doors=" doors
+                  "  floating=" floating "  unreach=" unreach))
+    (render-grid grid w h color?)
+    (println)))
+
+;; --- entry ---
+
+(let [w      (env-int "MAPGEN_W" 79)
```
nnunley: Can we actually use something like the edn file format for this? I know we can capture all environment variables in scripts, but it would make it slightly more data oriented.

nnunley (commented):
Generally looks good.

tools/mapgen.lg
mparrett: Sounds like a good idea, I'll look into it.

mparrett: Done — `mapgen` takes an EDN map now, inline or as a file path:

```
lg tools/mapgen.lg '{:seed 42 :depth 5}'
lg tools/mapgen.lg scenario.edn
```

merged over defaults. I kept the `MAPGEN_*` env vars as a fallback so existing recipes keep working, but the EDN argument wins. While in here I also pulled the audit metrics out of the script into `tools/mapaudit.lg`, so the same `floating`/`unreach` checks can back a regression test over a scenario set down the line.

mparrett: Happy to rip out the env vars if you prefer. Admittedly not necessary with e.g. `'{:seed 42 :depth 5}'` being the next best alternative.

nnunley: Yeah. I think I'd prefer that.

nnunley (approved):
Looks good to me, now.

PR #89, opened 2026-06-24
Guarantee level connectivity with union-find

mparrett:
Builds on #88 (the mapgen viewer + audit tool) — its `unreach=` metric is how this is measured, and the PR is based on that branch so the diff here is just the connectivity change. If #88 lands first, this retargets to `main`.

## The problem

`ensure-connectivity!` is best-effort. It loops up to 20 times; each iteration runs a full-grid BFS connectivity check (`is-connected?`) and, if the level is still disconnected, carves a corridor between two random room centers. After 20 attempts it gives up and returns whatever it has, connected or not.

In practice it often doesn't connect. Across seeds 42–91 at 79×30 depth 3, only 17 of 50 levels were fully connected; the rest had stranded floor (one seed left 37 tiles unreachable). Those are regions the player can see but can't walk to.

## The fix

Replace the retry loop with a disjoint-set (union-find) pass:

1. One O(W·H) sweep unions each floor tile with its right and down floor neighbor, labeling the connected components.
2. A second sweep carves one corridor from each distinct component to the first, unioning as it goes.

It never gives up, so the level is provably connected.

Complexity, for `n = W·H` (α is the inverse Ackermann function, ≤ 4 for any real grid):

- Before: up to 20 retries, each a full O(W·H) BFS → O(20·W·H) worst case per call, and still best-effort. Two call sites (after corridor carving, and after lakes/bridges) means up to ~40 full-grid floods per level.
- After: a single near-linear pass → O(W·H·α(n)) ≈ O(W·H), guaranteed.

After the change, 49 of those 50 seeds are fully connected.

## The one residual

Seed 48 still reports `unreach=1`, and it's a separate bug, not a connectivity miss. The floor is fully connected; the stranded tile is the up-stair, placed on a room center that a chasm floods, leaving the stair itself boxed in. Stairs are placed without a reachability check — a separate fix. I note it here because the audit catches it.

## Determinism and the test change

Still deterministic by seed. The new pass consumes no RNG and no longer threads through `det/with-context`, so a given seed's downstream layout differs from the old output. That shifted two ring tests in `container_trigger_test.lg` that pinned a seed's absolute player HP. They now assert the ring's effect as a delta against a no-ring run of the same world and action, which isolates the ring and holds across generator changes. The change also drops the now-dead `is-connected?` / `flood-count` BFS helpers.

## Verifying

```
for s in $(seq 42 91); do MAPGEN_SEED=$s MAPGEN_DEPTH=3 MAPGEN_COLOR=0 lg tools/mapgen.lg | head -1; done
```

Human-verified, `unreach=0` on every seed but 48 (the stair). `make test` is green.

## Captures
image

image

image

### DIFF
--- a/xsofy/terrain.lg
+++ b/xsofy/terrain.lg
@@ -324,60 +324,9 @@
             [(nth @carved mid) w3])
           [nil w3])))))
 
-;; --- Connectivity Check ---
-;; Array-based BFS flood fill — no atoms, no allocations in the hot loop.
-
+;; Floor set shared by the connectivity pass (ensure-connectivity-dsu!) and the
+;; door/feature phases.
 (def floor-tiles #{1 2 3 5 10 11 12 13 14 15 16 17 18 19})
-
-(defn flood-count [grid w h sx sy]
-  "Count reachable floor tiles from (sx,sy) using array-based BFS.
-   Marks visited before enqueueing so queue stays bounded to w*h."
-  (let [sz (* w h)
-        visited (int-array sz)
-        qx (int-array sz)
-        qy (int-array sz)
-        si (+ sx (* sy w))]
-    (if (or (< sx 0) (>= sx w) (< sy 0) (>= sy h)
-            (not (contains? floor-tiles (aget grid si))))
-      0
-      (do (aset visited si 1)
-          (aset qx 0 sx)
-          (aset qy 0 sy)
-          (loop [head 0 tail 1 cnt 1]
-            (if (>= head tail)
-              cnt
-              (let [cx (aget qx head) cy (aget qy head)
-                    t (loop [di 0 t tail]
-                        (if (>= di 4) t
-                          (let [nx (+ cx (aget dir-dx di)) ny (+ cy (aget dir-dy di))]
-                            (if (in-grid? w h nx ny)
-                              (let [ni (+ nx (* ny w))]
-                                (if (and (= (aget visited ni) 0)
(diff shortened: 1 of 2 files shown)


nnunley:
I think you might want to try a* based digging as an alternative - add rooms, and then have the A* algorithm add additional corridors and doors until they are strongly connected to the goal(s). But having union-find as a fast validation makes sense.

mparrett:
Thanks — I prototyped this. Kept union-find as the connectivity validator, as you suggested, and swapped only the connector dig to A*: same component detection, but each stray region routes to the connected mass via A* over a weighted grid (cheap to cross existing floor, costly to dig new) instead of the straight L-corridor.

It connects every level and stays deterministic. On the floating-door question, though, A* routing alone was a wash — across 200 levels (seeds 42–91 × depths 1/3/5/10) total floating held about flat (~3400 → ~3560) at the same connectivity. Re-routing the corridors doesn't move the count.

That lines up with the "corridors *and* doors" in your comment: floating doors come from the door-placement pass dropping doors on corridor tiles that aren't real thresholds, so moving the corridors just relocates them. The doors are the lever, not the routing.

The door-placement fix gets floating to 0 in our testing; it's up as #92. For connectivity itself we'll keep the straight-line dig: A* routing is appealing for more organic corridors, but it's orthogonal to the floating count, and a fair bit more code (a weighted pathfinder vs. two carves) for no measurable win here.

PR #90, opened 2026-06-24
Place stairs on a reachable floor tile

mparrett:
Builds on #89 (the connectivity fix): it clears the one `unreach=1` case #89 called out, and is based on that branch so the diff here is just the stair change. If #89 lands first, this retargets to `main`.

## The problem

Stairs are placed at the first and last room centers (`(first @rooms)` / `(last @rooms)`) with no check that the tile is passable. When a lake or chasm floods that center, the stair lands in hazard and is stranded from the rest of the level. On seed 48 the up-stair ends up walled in by chasm, leaving `unreach=1` even after the connectivity fix.

## The fix

Pick the first (or last) room center that is dry floor (tile 1 or 2), falling back to the first/last room if none qualify. Stairs are already the last mutation and run after the final connectivity repair, so a floor room center is reachable.

No RNG is consumed (only the stair tile moves), so this changes stair placement and nothing else downstream.

## Result

`unreach=0` across seeds 42–91 at depth 3 and 42–71 at depth 6. Together with #89, every floor in that range is fully connected, stairs included.

## Verifying

```
for s in $(seq 42 91); do MAPGEN_SEED=$s MAPGEN_DEPTH=3 MAPGEN_COLOR=0 lg tools/mapgen.lg | head -1; done
```

Every seed reports `unreach=0`. `make test` is green.

## Screenshot (seed 48 unreach=0)

without this PR
image

with this PR
image

### DIFF
--- a/xsofy/terrain.lg
+++ b/xsofy/terrain.lg
@@ -800,6 +800,24 @@
         (swap! rooms conj [(quot (+ x1 x2) 2) (quot (+ y1 y2) 2)])
         w6))))
 
+(defn stair-tile
+  "First (or last, if reversed?) room center that is a passable floor tile
+   (1/2), so a stair never lands on a lake- or chasm-carved hazard. Stairs are
+   the last mutation and run after the final connectivity repair, so a floor
+   room center is reachable. Falls back to the first/last room if none qualify."
+  [grid w rooms reversed?]
+  (let [v (vec rooms)
+        n (count v)]
+    (loop [k 0]
+      (if (>= k n)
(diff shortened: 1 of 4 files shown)


mparrett:
The changes from this PR didn't make it to `main`. The stack was squash-merged bottom-up — #89 merged into #88's branch before #90 was added on top of #89 — so #88's squash carried the viewer, audit, and connectivity, but not this PR's stair fix and regression test. Re-submitted against current `main` in #91.

PR #91, opened 2026-06-24
fix(terrain): place stairs on a reachable floor tile (+ regression test)

mparrett:
Stairs are placed at room centers, but a room center can be carved into a lake or chasm, or otherwise sit off the connected floor. When that happens the stair lands on a tile that isn't reachable. Across a 200-level matrix (seeds 42–91 × depths 1/3/5/10) on current `main`, 6 levels have an unreachable tile for this reason.

`stair-tile` walks the room centers and picks the first (or last) one that is a passable floor tile, falling back to the first/last room if none qualify. Stairs are the last mutation and run after the final connectivity repair, so a floor room center is reachable. With the fix the same matrix is clean — `unreach=0` everywhere.

The second commit adds a regression test that locks it. It loads a scenario matrix (`test/fixtures/dungeon-scenarios.edn`) and asserts `tools.mapaudit/audit` over each generated level: `unreach=0` hard, plus floating-door count under a loose ceiling. The ceiling is a coarse tripwire, not a cleanliness target: floating doors are a known separate issue (3 to 42 per level), so the test only catches a generator that drowns levels in them. The contract lives in the EDN, so tightening it later is a fixture edit rather than a code change.

## Why a new PR

This is the stair fix and test from #90, which GitHub shows as merged but whose content never reached `main`. The #88 → #89 → #90 stack was squash-merged bottom-up within a few minutes: #89 merged into #88's branch first, then #90 merged into #89's branch afterward, so by the time #88 squashed into `main` it carried the viewer, audit, and connectivity work, but not #90's two commits. #90 reads as merged because its head merged into its base, but its changes stopped there. Re-submitting them against current `main` as a minimal two-commit change.

## Before/after

image

image

### DIFF
--- a/xsofy/terrain.lg
+++ b/xsofy/terrain.lg
@@ -800,6 +800,24 @@
         (swap! rooms conj [(quot (+ x1 x2) 2) (quot (+ y1 y2) 2)])
         w6))))
 
+(defn stair-tile
+  "First (or last, if reversed?) room center that is a passable floor tile
+   (1/2), so a stair never lands on a lake- or chasm-carved hazard. Stairs are
+   the last mutation and run after the final connectivity repair, so a floor
+   room center is reachable. Falls back to the first/last room if none qualify."
+  [grid w rooms reversed?]
+  (let [v (vec rooms)
+        n (count v)]
+    (loop [k 0]
+      (if (>= k n)
+        (if reversed? (last v) (first v))
+        (let [idx (if reversed? (- n 1 k) k)
+              [x y] (nth v idx)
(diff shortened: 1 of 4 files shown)


mparrett:
> This is the stair fix and test from https://github.com/nooga/xsofy/pull/90, which GitHub shows as merged but whose content never reached main.

@nnunley Not sure if this was a local mix-up + claude confusion; juggling too much today. In any case, thanks for the reviews on all of this!

nooga (approved):

nooga (approved):

PR #93, opened 2026-06-25
xsofy workbench: static, server, and wasm

mparrett:
## xsofy workbench: static, server, and wasm

A tight `edit generator code → regenerate → see + measure` loop for dungeon terrain work. It answers "did this change improve connectivity / feel / cost" without a rebuild between iterations — the signal the substrate/connector proposals need cheaply. This folds the exploded 3D viewer in: its render-only view is now the workbench's **static mode**, so there's one page and one renderer instead of two near-identical copies.

### One UI, three modes

Rendering and controls sit behind a single `Backend` interface (generate floors, eval code). Each backend declares its capabilities (`live`, `eval`, `upload`) and the UI shows or hides controls off those — no per-mode branches. The page auto-detects which mode to run, and `?mode=` forces it:

- **static**: render a prebuilt export (an embedded one, a default served alongside the page, or a dropped `.json/.js`). No engine, no server, no cross-origin isolation: the lowest-overhead path. It hosts statically and opens offline, for sharing a snapshot.
- **server**: `workbench-server.lg`, a localhost lg HTTP server. `GET /floors` generates; `POST /eval` (and an optional nREPL) redefine generator fns in the live `xsofy.terrain` image, so the next `/floors` picks them up. Native-speed gen, persistent image, reset = restart. The fastest local loop.
- **wasm**: the let-go image runs in the page via `lg -w -w-host-eval`, driven through the public `LetGoHost.eval`. Reload = fresh VM, and it hosts statically for a shareable live playground. Same UI, same `dungeon-build.lg` floors+metrics builder; only a thin `WasmBackend` adapter differs.

Detection is probe-free: `window.LetGoHost` (present in a `-w-host-eval` bundle) selects wasm; a marker the server injects into its served page selects server; neither selects static, the zero-dependency default. The mode pill in the panel names the active mode; clicking it explains that mode and switches to the others the page can actually back. Asking for a mode the page can't serve (e.g. `?mode=wasm` on a plain static host) falls back with a notice rather than failing silently.

### Metrics per floor

Surfaced alongside each generated floor, as regression signals for connectivity work: **components** (1 = the reachability invariant holds), **dead-ends**, **rooms**, **gen-ms** (per-floor wall time), and **loops**. `loops` is a coarse cyclomatic count over the raw tile graph — a placeholder that over-counts open rooms; the real "designed loop" signal wants a room/junction graph.

### Design notes

- **Eval results come back out of band via `js/emit`, not the eval return value.** `LetGoHost.eval` hands back a stringified value, and a returned JSON string would arrive double-quoted (let-go's `String` prints quoted). So `generate` and `eval-str` emit their payloads (`xsofy/floors`, `xsofy/eval-result`); the UI reads those events. This also keeps the backends shape-identical.
- **The wasm-mode page is injected into the host-core bundle, so it shares global scope with it.** A bare top-level `const` that matches a host-core name (e.g. `WASM_MODE`) is a redeclare error that silently kills the whole script. The workbench script is wrapped in an IIFE to scope it.
- **`WasmBackend` serializes its calls.** Overlapping generations would otherwise each arm a listener for the same emit, and the first event would resolve all of them — a newer request rendering an older floor set. One call in flight at a time keeps each listener matched to its own emit.

### let-go dependency

Wasm mode needs a let-go built with `-w-host-eval` (shipped in v1.11.0). Server and static modes don't. `README.md` has the build steps.

### Security

`/eval` executes arbitrary code by design (it's the live-coding REPL), so it's locked to same-origin: the server binds to `127.0.0.1`, sends no `Access-Control-Allow-Origin`, and rejects any request whose `Origin` isn't its own — a page on another origin can't POST code to it. Requests with no `Origin` (curl, local tooling) still pass. It stays a dev-only tool.

### Testing

All three modes validated in a browser: static renders an export with the live controls and eval console hidden; server and wasm generate live and redefine into the resident `xsofy.terrain` image. The change is confined to the `tools/workbench` tooling — no game source touched.

**wasm mode**
image

**static mode**
image

**server mode**
image


mparrett:
Self-review findings:

**High: generated floor data is stale.** Re-running the PR's exporter does not reproduce any checked-in data file: `tools/dungeon-3d/floors.js`, `tools/dungeon-3d/floors-seed64.json`, and `tools/dungeon-3d/floors-seed7-d8.json` all differ from fresh output. That means following the documented generation commands will create unrelated churn, and the bundled viewer data is not traceable to the exporter in this PR. I verified this with the repo's `lg` binary; all three comparisons returned `DIFF`.

**Medium: the exporter does not use the same seed derivation as the game floor path.** `tools/dungeon-3d/export-floors.lg` calls `terrain/generate-level` with raw `{:seed SEED}`, while game floor creation derives `floor-seed` with `det/combine seed-input [:floor depth]` before generating terrain in `xsofy/world.lg`. So the README's "what the game generates" claim is misleading unless this tool intentionally uses terrain seeds rather than game seeds.

Checks run: checked out PR 93 in a temp clone, loaded the viewer via localhost with Playwright, captured a screenshot, clicked `top`, opened `view & tuning`, parsed all bundled exports, and ran the exporter. The UI booted and basic controls worked; the main issues are data reproducibility and seed semantics.

mparrett:
Both addressed in f518c58.

**Stale data (High).** Regenerated `floors.js`, `floors-seed64.json`, and
`floors-seed7-d8.json` against this branch's `terrain.lg` with the fixed exporter.
The bundled data had been captured against a diverged `terrain.lg`, which is why it
didn't reproduce. The documented generation commands now reproduce the bundled
files byte-for-byte, and the exporter is deterministic across repeated runs.

**Seed semantics (Medium).** `export-floors.lg` now derives each floor's layout
seed with `(det/combine SEED [:floor depth])`, mirroring `xsofy.world/generate-floor`,
instead of passing the raw origin seed. A given seed now reproduces the floors the
game actually builds, so the "what the game generates" claim holds; the docstring
says as much.

Re-loaded the viewer on the regenerated data: 5 floors render, stairs present per
floor, no console errors.

mparrett:
A note on scope, since the tool is intentionally minimal.

**Runtime requirement.** `export-floors.lg` needs let-go `main`: native xxh3 (`xsofy.det`) plus `core/*command-line-args*` (#205), which isn't in any tagged release yet. So the tool tracks `main` rather than pinning a version; on too-old lg it fails with a clear `*command-line-args*` resolve error naming the line. The viewer (`index.html`) has no runtime dependency — plain JS/CSS over a `<script>`-loaded data file.

**Why static export rather than live in-browser generation:**

- Single self-contained `index.html` — opens from `file://`, no server, no cross-origin isolation, no wasm payload.
- The bundled data reproduces byte-for-byte from the documented commands.
- A live-wasm path (generate in-browser for dynamic seed/parameter refresh) is a clean follow-up, not a blocker: the `js/emit` + `Eval` host bridge it needs is already upstream. What keeps it out of scope here is the wasm bundle's size and boot cost, plus that dynamic gen-parameter sliders would want a `generate-level` overrides hook that doesn't exist yet.

Static export covers "look at real generated dungeons" with none of that overhead.

nooga (approved):

nooga:
I can see this becoming a workbench for the terrain gen!

mparrett:
Review findings from current head (`e9e2e817a9001569d4faedb7a132781fb48e9fd3`):

**P2: wasm mode is detected too broadly.** `tools/workbench/workbench.html` treats any `window.LetGoHost` as wasm-capable, but let-go exposes `LetGoHost` for normal wasm bundles too; only `LetGoHost.eval` is gated by `-w-host-eval`. A non-host-eval wasm page will select wasm mode, then fail at generation with `LetGoHost.eval is not a function` instead of falling back. Check `typeof window.LetGoHost?.eval === "function"` for wasm capability.

**P3: server mode advertises static switching but does not serve the default export.** `tools/workbench/workbench-server.lg` returns `workbench.html` for every non-API path, so `GET /floors-seed7-d8.json` returns HTML with `200`. When switching to `?mode=static` from the server page, `workbench.html` tries to parse that as JSON and falls back to the upload prompt. Either serve the bundled `.json` files with JSON content types or do not present static as fully backable from the server page.

Checks run:
- Fetched PR 93 at head `e9e2e817a9001569d4faedb7a132781fb48e9fd3`.
- Re-ran all documented exports; `floors.js`, `floors-seed64.json`, and `floors-seed7-d8.json` reproduce byte-for-byte.
- Ran `dungeon-build` via local let-go binary.
- Started `workbench-server.lg` on `127.0.0.1:7077`; verified `/floors` works and cross-origin `/eval` is rejected.

mparrett:
Addressed all findings.

PR #105, opened 2026-06-30
fix(wasm): keep the COI service worker when it's the isolation source

mparrett:
## What

`patch_wasm_coi.lg` injects a startup guard: once the page is `crossOriginIsolated`, it unregisters all service workers as cleanup. That cleanup is unconditional, and on GitHub Pages it's self-defeating.

Pages can't send COOP/COEP headers, so the service-worker shim is the *only* thing that makes the page `crossOriginIsolated`. So the sequence on Pages is: register the SW, reload, the SW now controls the page and isolation flips true, then that cleanup unregisters the SW that just provided it. Next visit there's no SW, so it registers and reloads again — a re-register + reload on every visit. And `getRegistrations()` is origin-wide, so a preview reaching isolation also tears down the root deploy's SW (and vice versa) once multiple deployments share the origin.

## The fix

Gate the unregister on `!navigator.serviceWorker.controller`:

- No controller → isolation came from real COOP/COEP headers, the SW is redundant, drop it (unchanged behavior on a header-isolated server).
- A controller → the SW is the isolation source (the Pages case), keep it.

The retry-param cleanup still runs whenever isolated, so the `_lg_coi_retry` URL param is still stripped.

## Impact

On the current single-site deploy this removes a reload on every visit. It also makes per-deployment service workers coexist (root and any `/pr-N/` previews stop tearing down each other's SWs), which the branch-publishing work needs.

## Verification

Built a bundle with the patched script and deployed it to a real GitHub Pages preview. On load the page reaches `crossOriginIsolated` via the SW (controller set), and the SW stays registered — `getRegistrations()` is 1, not 0. A second load is already isolated with the SW intact and no `_lg_coi_retry` round-trip, i.e. no re-register churn. The generated guard is `if (!navigator.serviceWorker.controller)` around the unregister, with the retry-param cleanup outside it.

## Scope

One guard in `patch_wasm_coi.lg`; no other changes. Behavior on a header-isolated server is unchanged.


mparrett:
A correctness check before this merges: does the game rely on the every-visit reload the current code does? It doesn't.

Test: two visits in a persistent browser profile, so the service worker registered on the first visit survives to the second. The second visit is the no-reload path — the SW already controls the page, so isolation is present without the register+reload dance.

| | Visit 1 (cold) | Visit 2 (SW controls, no reload) |
|---|---|---|
| `crossOriginIsolated` | true | true |
| `_lg_coi_retry` reload | the install dance | none |
| booted to title | — | yes |
| keyboard input advanced the game | — | yes (22 rows of gameplay) |

Visit 2 is the one that matters: it loaded isolated with no reload, the game booted, and input worked. So `SharedArrayBuffer` (the `read-key` ring) is live when isolation comes from a pre-existing SW controller, not only from a just-completed reload. The reload was overhead, not something the game depends on to settle state.

The fallback is intact: the register+reload branch is untouched, so any load that isn't isolated (first-ever visit, evicted SW, hard cache clear) still reloads exactly as before. The guard only skips the reload when isolation is already present; it can't leave a page un-isolated.

Ran on a Pages deploy of the generic-shell build, but the guard is shell-independent, so it holds for the shipping xsofy-shell bundle.

nooga (approved):

mparrett:
Post-merge review:

No blocking findings; verified:

- Pinned v1.11.1 WASM build and patch execution succeed.
- No-header browser boot reaches isolation and renders correctly.
- Reload retains the clean URL and avoids retry churn.
- Both PR checks pass.

Residual gap: CI builds WASM but does not execute `patch_wasm_coi.lg` or test the service-worker bootstrap path; verification currently depends on manual coverage.

PR #108, opened 2026-07-01
perf: gate :lights cache invalidation on real terrain mutation

mparrett:
`step-fire` set `:lights nil` every turn unconditionally, forcing `collect-terrain-lights` into a full O(w·h) terrain scan on the next `update-fov`. The cache only needs invalidating when a light-emitting tile is created or removed (fire ignited / burned out to ash). Now gated on actual terrain mutation.

**Scope changed from the original PR:** the `20b4e36` wander offset-scan commit is **dropped** — its random-offset scan biased wandering direction (measured ~8.7k east vs ~1.2k west over 10k seeds). This PR is now the lights-cache gate only.

**Review P1 addressed** (`3f57a9a`): the gate initially only tracked mutations inside `step-fire`, so ignitions from `burning-entities-ignite` / dragon / rune fire (via `ignite` / `ignite-forced`) left the cache stale — a freshly ignited fire emitted no light until it spread. Now the ignite helpers invalidate `:lights` themselves, on every call site, only when fire is actually placed. Added `xsofy/test/fire-test` asserting ignite/ignite-forced invalidate and a no-op ignite preserves the cache.

xsofy suite 282/2730 green; native compile clean.

### DIFF
--- a/xsofy/fire.lg
+++ b/xsofy/fire.lg
@@ -19,25 +19,34 @@
 (def fire-default-ttl 5)
 (def protected-tiles #{13 14})  ;; stairs must survive dragon fire and rune fire
 
+;; Fire is light-emitting, so any ignition must invalidate the :lights
+;; cache — step-fire's own terrain-changed gate only sees fires it spreads
+;; or burns out, not ignitions from burning-entities-ignite / dragon /
+;; rune fire, which run in separate phases. Invalidating here (not just in
+;; step-fire) keeps the cache correct for every call site.
 (defn ignite [world x y]
   "Set a tile on fire. Only works on flammable tiles or if forced."
   (let [{:keys [terrain width]} world
         tile (terrain/tget terrain width x y)]
     (if (or (contains? terrain/flammable-tiles tile) (= tile 1))
       (do (terrain/tset! terrain width x y fire-tile)
-          (assoc-in world [:fire-ttl [x y]] fire-default-ttl))
+          (-> world
+              (assoc-in [:fire-ttl [x y]] fire-default-ttl)
+              (assoc :lights nil)))
       world)))
 
 (defn ignite-forced [world x y ttl]
   "Force fire at a position regardless of terrain."
   (let [{:keys [terrain width]} world
-        tile (terrain/tget terrain width x y)]
-    (when (and (not= tile 0) (not= tile 8) (not= tile 7)
-               (not (contains? protected-tiles tile)))  ;; not void, wall, lava, stairs
+        tile (terrain/tget terrain width x y)
+        placed? (and (not= tile 0) (not= tile 8) (not= tile 7)
+                     (not (contains? protected-tiles tile)))]  ;; not void, wall, lava, stairs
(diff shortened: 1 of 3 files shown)


nooga (approved):

mparrett:
I found two issues that should be addressed before merge:

1. **[P1] Ignition can leave the `:lights` cache stale.** `step-fire` now preserves `:lights` when it sees no mutation of its own, but `burning-entities-ignite` runs afterward and both `ignite` and `ignite-forced` mutate light-emitting terrain without invalidating the cache. A newly ignited fire can therefore emit no light until it spreads or burns out. I reproduced this with a burning entity on grass: the tile became 15 while cached lights remained `[]`, although `collect-terrain-lights` returned the expected fire light. Please invalidate in the ignition helpers or after all fire-mutation phases.

2. **[P2] The offset scan makes wandering directionally biased.** A random starting offset followed by the next legal entry in fixed `dirs8` order is not uniform among legal moves. In a horizontal corridor where only east and west are legal, east occupies 7/8 of the starting offsets and west only 1/8. Across 10,000 deterministic seeds I reproduced 8,765 east moves versus 1,235 west moves. The previous Fisher-Yates implementation selected the legal directions uniformly. Please choose uniformly from the valid candidates, or explicitly accept and test this gameplay behavior change.

CI is green, but neither regression is asserted by the current suite.

mparrett:
Thanks — both addressed:

- **P1 (stale cache on ignition):** fixed in `3f57a9a`. `ignite` and `ignite-forced` now invalidate `:lights` directly, so `burning-entities-ignite` and dragon/rune fire are covered, not just `step-fire`'s own spread/burn-out. Invalidation is gated on fire actually being placed, so a no-op ignite still preserves the cache. Added `xsofy/test/fire-test` with the ignite/ignite-forced/no-op cases (your repro is now asserted).
- **P2 (wander bias):** dropped the `20b4e36` offset-scan commit entirely. The speedup came from stopping at the first legal direction, which is what broke uniformity — not worth the gameplay change here. This PR is now the lights-cache gate only.

mparrett:
Note: before merging, @mparrett Please quantify/measure the performance improvement; need to understand if this optimization actually earns its keep. (quite possible, on mobile devices)

mparrett:
Ran a native A/B on this (same lg binary, rebased head vs origin/main). The averaged engine fixtures land within noise, which undersells the change: the saving is real, but small enough that run-to-run variance swamps it.

The measurement that isolates it: `collect-terrain-lights` costs ~2ms/call on a 79×30 floor (timed alone, ~5k calls/seed). Baseline's `step-fire` sets `:lights nil` every turn, so it pays that O(w×h) scan on every turn regardless of fire activity. Gating on real terrain mutation drops it from settled turns, which are the common case once a floor is quiet. That's roughly 10% of a 15–20ms turn in steady state.

Fire-heavy turns are the break-even case (both branches invalidate), and showed no meaningful regression from the added atom/branch. So the win is steady-state per-turn, not a fire-scenario spike.

PR #109, opened 2026-07-01
fix(ui): death-screen log — drop redundant death line + pad scroll residue

mparrett:
Two death-screen message-log fixes that are on the daily branch but not on `main`.

**Redundant "You die…" line (`world.lg`)** — when death follows a kill, combat already logs "The X kills you!", and the death branch then also logs "You die…", so the log shows both. Drops the redundant "You die…" when a cause-of-death line was already logged that turn.

**Message-log scroll residue (`render.lg`)** — the death-screen log rows were written without right-padding, so scrolling up left tails of longer earlier lines behind shorter ones. Pads each rendered row to width.

Both touch only the death-screen log. xsofy suite 279/2725 green; native compile clean.

Companion to the `d3ac676` re-clear-on-resize fix, which needs a manual rebase over the merged #101 render-death-screen change and will follow in a separate death-screen-polish PR.

### DIFF
--- a/xsofy/render.lg
+++ b/xsofy/render.lg
@@ -1005,7 +1005,7 @@
       (term/move-cursor 2 log-top)
       (term/set-fg 90 80 65)
       (term/set-bg 10 10 10)
-      (term/write (str "─── Messages " (if (> max-offset 0) (str "(" (- log-count offset) "/" log-count ") ") "") "───"))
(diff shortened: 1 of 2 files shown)


nooga (approved):

mparrett:
Tested- intended effect confirmed.

mparrett:
Clean.

image

PR #111, opened 2026-07-02
Native engine benchmark + shared xsofy.perf core

mparrett:
# Native engine benchmark + shared `xsofy.perf` core

## Why

We have no way to measure the engine's per-turn cost on its own. Every timing
number today is entangled with the browser: WASM, the JS bridge, and the xterm
canvas all sit between a turn and the clock. That makes "is the game slow?" and
"is the *shell* slow?" the same unanswerable question.

This adds a native, headless baseline — Tier 1. It runs the same `lg` binary
that runs the game, headless: no render, no bridge, no terminal. It times
`update-world` directly. Once the in-bundle WASM harness lands and drives the
same fixtures, the browser overhead falls out as a straight delta instead of a
guess.

## What

- `xsofy/perf.lg` — a UI-agnostic timing core: `time-turn`, `run-trial`,
  `percentile`, `summarize`, `run-fixture`. World and input only.
- `bench/native.lg` — a native CLI driver (`lg bench/native.lg`) that defines
  the fixtures and prints the table.

Four fixtures, chosen to exercise distinct cost centers: `:wait` (idle turn —
AI + floor effects with the player still), `:autoex` (BFS pathfinder + full
render cadence), `:descend` (one descent then idle — the floor-gen + entity/lighting/fire spike), and
`:busy` (long autoexplore for allocator behavior).

```
fixture           n    avg    p50    p95    max
--------     ------ ------ ------ ------ ------
:wait           300    10ms    10ms    18ms    23ms
:autoex         300    11ms    11ms    15ms    18ms
:descend        300    11ms    11ms    17ms    56ms
:busy          1002     9ms    10ms    13ms    17ms
```

## Design notes

- **The core is deliberately dependency-free** (no render/bridge/xterm/println).
  That is what lets the native harness stand alone, and what lets the future
  WASM harness drive the *same* core and fixtures so the two numbers compare
  directly. The drivers own presentation and seed policy; the core owns timing.
- **Seed = trial index.** Each trial samples a distinct world, so averaging
  across the spread also smooths timing noise (GC, scheduling). Determinism
  comes from `make-world` threading the seed through the world's RNG and
  `:next-id`, so a given seed replays bit-identically. `run-fixture` takes
  `(seeds, trials)` explicitly, so a caller can instead pin one world and repeat
  it — the one real policy difference between drivers is now an argument, not a
  fork.
- **No `(ns …)` form in `bench/native.lg`.** `lg` evaluates a namespaced file
  twice (script eval + namespace load), which would run the bench twice. The
  bare `require` + `alias` is the same idiom `tests/run.lg` uses.

This is purely additive: `main` has no bench today.

### DIFF
--- a/bench/native.lg
+++ b/bench/native.lg
@@ -1,0 +1,47 @@
+;; bench/native.lg — native engine benchmark (Tier 1).
+;;
+;; Drives xsofy.perf's shared core against the same lg binary that runs the
+;; game, headlessly: no render, no bridge, no xterm. Measures the engine's
+;; per-turn cost in isolation so later overheads (WASM/JS bridge/xterm canvas)
+;; can be attributed by delta. The in-bundle WASM harness (xsofy/bench.lg)
+;; drives the same core + fixtures so the two numbers diff directly.
+;;
+;; Run:  lg bench/native.lg
+;;
+;; Seed strategy: seed = trial index, so each trial samples a distinct world;
+;; averaging across the spread also smooths timing noise (GC, scheduling).
+;; rand-int is cosmetic-only and never touches gameplay state.
+;;
+;; No (ns ...) form: lg evaluates a namespaced file twice (script + namespace
+;; load), which would double-run the bench. tests/run.lg uses the same trick;
+;; the bare-symbol require + separate alias below is the non-ns idiom.
+
+(require 'xsofy.perf)
+(alias 'perf 'xsofy.perf)
+
+(def fixtures
+  [;; baseline: entity AI + floor effects tick, player idle — "what does an
+   ;; empty turn cost?"
+   [:wait    (fn [_] ".") 60]
+   ;; autoexplore: BFS pathfinder + render-full cadence as the camera shifts.
+   [:autoex  (fn [n] (if (zero? n) "x" ".")) 60]
+   ;; descend-then-wait: reach deeper floors (more entities/lighting/fire).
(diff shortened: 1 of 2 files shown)


mparrett:
Review finding — **P2: empty benchmark samples crash `summarize`.**

`xsofy/perf.lg` calls `percentile` before handling the empty case, so `summarize []` raises `nth index out of bounds` (reproduced locally). `run-fixture` can produce an empty aggregate with empty seeds, zero trials, or a zero-turn fixture. Please short-circuit the empty case or make `percentile` return a defined default.

Relevant code: https://github.com/nooga/xsofy/blob/590f27e9e1489161185e9d3193020930d939b330/xsofy/perf.lg#L40-L55

mparrett:
Confirmed and fixed (folded into the commit, force-pushed to `b2e8f14`).

`summarize` now short-circuits the empty case and returns `{:n 0 :avg 0 :p50 0
:p95 0 :max 0}` before it ever calls `percentile`, so the `(nth sorted -1)` on
an empty vector is gone. The non-empty branch keeps the existing behavior.

Checked the three paths you named — empty seeds, zero trials, and a zero-turn
fixture all return the zero summary now instead of raising; a populated run is
unchanged (`[10 20 30 40]` → avg 25, p50 20, p95 30, max 40).

mparrett:

PR #112, opened 2026-07-02
Responsive camera + narrow-terminal rendering

mparrett:
## Why

The renderer assumes the terminal is big enough to show the whole map. On a small viewport — a phone in portrait, or a narrow native TTY — two things break:

- World cells that fall outside the visible rect are silently dropped, including the player. You can walk off the edge of what's drawn.
- A fixed 20-column sidebar eats a third of a narrow screen, leaving a useless strip of actual map.

This makes rendering responsive to the viewport it's given: a camera that follows the player, a layout that collapses chrome when space is tight, and dead-zone margins that hold a constant feel across sizes.

## What

Five commits, `render.lg` + `ui.lg`, +296/-110:

1. **Dead-zone camera + viewport follow.** A camera translates world coordinates to terminal cells and follows the player within a dead zone.
2. **Narrow-terminal layout collapse.** Sidebar, log, and status hint drop out below width/height thresholds; the map takes the freed space.
3. **Proportional dead-zone margins.** Per-axis percentage of the viewport instead of a fixed cell count.
4. **Visible-window render fix (`2c29609`).** `render-map` walks the visible camera window and translates back to world coordinates instead of iterating from the world origin — closes the terrain gaps that showed on any scrolled map.
5. **Comment fix (`cfe6f63`).** Docs-only.

## Design notes

- **`map-cell` is the linchpin.** It translates world `(x y)` to terminal `[col row]` through the camera offset and returns `nil` for anything off the visible window. Every renderer (`render-map-tile`, `render-entity`, and the VFX frames) goes through it, so culling lives in one place instead of scattered inline bounds checks. `render-dirty` compares the last committed camera to the current one: on a camera move it does a full render, otherwise it keeps the FOV-delta fast path.
- **The layout collapse is a layout decision, not a feature loss.** Below the thresholds each panel returns `nil` from layout and its renderer no-ops. The HP/depth/quest info the sidebar carried is already duplicated by the HTML host's info bar under WASM, and the same thresholds give small native TTYs the same treatment.
- **Margins are proportional because cells aren't square.** Terminal cells are roughly 2.3× taller than wide, so an equal cell count per axis gives unequal visual look-ahead; a fixed count is also a different *fraction* of the viewport at every size, so the feel drifted on resize. Per-axis percentages (`deadzone-pct-x` 60 / `deadzone-pct-y` 35), capped at half the viewport, preserve the no-oscillation guard and let the axes tune independently.

Applies to both the native TUI and the browser/WASM build — same render path.

### DIFF
--- a/main.lg
+++ b/main.lg
@@ -326,6 +326,9 @@
    gameplay seed chain — so this world replays identically to one made by
    new-game from the same seed."
   [seed title scheme quest]
+  ;; Fresh run: drop any prior run's camera so a same-depth/same-viewport
+  ;; restart re-centers on the player instead of inheriting the old offset.
+  (render/reset-camera!)
   (-> (world/make-world 79 30 seed)
       (assoc :title title
              :scheme scheme
--- a/xsofy/render.lg
+++ b/xsofy/render.lg
@@ -374,8 +374,9 @@
         (render-status-bars r (inc row) c)))))
 
 (defn render-sidebar [world r]
-  (ui/clear-rect r)
-  (when-let [player (get-in world [:entities :player])]
+  (when r
+    (ui/clear-rect r)
+    (when-let [player (get-in world [:entities :player])]
     (let [hp (or (get-in player [:body :hp]) 0)
           max-hp (or (get-in player [:body :max-hp]) 1)
           str_ (or (get-in player [:body :strength]) 0)
@@ -437,7 +438,7 @@
         (doseq [c creatures]
           (when (< @row (:h r))
             (let [next-row (render-creature-entry r @row c)]
-              (reset! row (or next-row (+ @row 2))))))))))
+              (reset! row (or next-row (+ @row 2)))))))))))
 
 ;; --- Log Panel ---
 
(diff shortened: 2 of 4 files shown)


mparrett:
Review finding — **P1: full renders skip the visible terrain once the camera offset is non-zero.**

`render-map` still iterates world coordinates from `(0,0)` through the viewport width and height, while `map-cell` now subtracts `:cx/:cy`. With a narrow viewport, the loop therefore never visits much or all of the visible world window `[cx, cx+w) × [cy, cy+h)`. Entities still render because their renderer iterates the full entity set, which can leave a screen containing only `@` and no terrain; I reproduced that at a 200 px viewport.

Please iterate from the camera offsets through the visible world bounds, or iterate screen cells and translate them back to world coordinates.

Relevant code: https://github.com/nooga/xsofy/blob/6683b264c9c76dcb907d27f8c764dd2e3d3ddda9/xsofy/render.lg#L254-L266

mparrett:
Confirmed and fixed in `2c29609`.

`render-map` walked world coordinates from `(0,0)`, so once the camera offset
was non-zero it iterated `[0, w)` while the on-screen window was `[cx, cx+w)` —
missing the visible cells entirely once `cx >= w`. It now walks screen cells and
translates back to world coordinates (`x = cx + sx`), covering exactly the
visible window clamped to the map bounds. `map-cell` and `render-map-tile` were
already correct; only the iteration range was wrong.

Reproduced your case at a 400px viewport: the map was blank but for `@` and one
adjacent entity before; a full room (walls, floor, and door) renders around `@`
after. Test suite green.

mparrett:
## Review findings

- **P1 — narrow native terminals lose critical HUD data.** At widths below 80, `layout` returns no sidebar (`xsofy/ui.lg:133–156`). That removes HP, status effects, depth, equipment, and visible-enemy information; the HTML fallback only exists under WASM. PR #119 fixes this, so #112 should not ship independently—merge them atomically or incorporate the top bar here.

- **P2 — camera state leaks into a new run.** Camera validity (`xsofy/render.lg:118–124`) keys only on depth and viewport dimensions. Restarting after dying on depth 1 therefore reuses the previous run’s camera. The restart paths call `render-full` without resetting it (`main.lg:474–477`, `main.lg:514–522`).

  Reproduction on the exact PR head:

  ```
  old camera: [59 10]
  new run with inherited state: [36 10]
  new run from clean state: [30 10]
  ```

  Reset camera state when constructing/restarting a run, with a regression test.

### Verdict

Request changes for the camera leak; require #119 as a release dependency.

Validation: exact head `cfe6f63`; 2,725 assertions and runtime smoke passed. No camera/layout-specific tests were added.

mparrett:

PR #113, opened 2026-07-02
Mobile touch shell + browser HUD bridge

mparrett:
# Mobile touch shell + browser HUD bridge

## Why

The published site had no touch controls. The mobile key-set (movement pad,
action buttons, and item/toggle panes) lived only in a dev shell that never
deploys, so on a phone there was nothing to tap. Separately, the browser HUD's
info bar had no data path from the game.

This stack adds the browser-side mobile ergonomics: a named-channel bridge from
the game to the host, the callers that feed it, and the touch UI plus a
populated info bar in the shell the deploy actually ships.

## What

Three commits, stacked:

1. **`ui_bridge.lg`** — wraps let-go's `js/emit` into named channels
   (`emit-stats`, `emit-startup`, plus custom channels). New file, no callers
   yet.
2. **Wire the callers** — `emit-startup` in `assemble-world` (the shared construction path, so replays populate the HUD too), `emit-stats` at the
   game-loop entry and after every `:game` turn (`main.lg`, `play.lg`).
3. **Touch UI in the shipping shell** — the movement pad, action column, and
   swappable MOVE/ITEMS/KEYS panes, plus the info bar the bridge feeds
   (`tools/xsofy-shell.html`).

## Design notes

- **The bridge is inert on native.** `js/emit` only exists under WASM, so the
  game calls `emit-*` unconditionally and native builds see a no-op. The bridge
  and its callers apply to all platforms but cost nothing in the TUI.
- **One shell, not two.** The touch UI moves *into* `tools/xsofy-shell.html`,
  the same shell the Pages deploy injects via `tools/inject_shell.lg`, rather
  than adding a parallel copy. It keeps that shell's correct core: the inlined
  base64 full-glyph font (no `unicode-range`, so xterm measures a single cell
  advance and the grid can't drift) and the `onReady` handler that awaits the
  font before `term.open()`. Dev-only affordances stay out of the shipping
  build.
- **The info bar is WASM-only.** It's populated by the emit channels the first
  two commits add, so it fills in under the browser build and is absent on
  native.
- **Independent of the responsive-render slice.** This stack works standalone;
  the in-terminal layout collapse on narrow viewports comes from the separate
  render PR. The two compose at integration — reviewing this one, the terminal
  keeps its full-width layout.

### DIFF
--- a/.github/actions/build-wasm/action.yml
+++ b/.github/actions/build-wasm/action.yml
@@ -52,15 +52,4 @@ runs:
     #    from 404-ing — the smoke gate counts any console error as a failure.
     - name: Write build-info.json
       shell: bash
-      run: |
-        letgo="unknown"
-        if [ -f .let-go-version ]; then
-          letgo="$(awk '!/^[[:space:]]*(#|$)/ {print $1; exit}' .let-go-version)"
-        fi
-        cat > "${{ inputs.dist }}/build-info.json" <<JSON
-        {
-          "xsofy": "$(git rev-parse HEAD)",
-          "let-go": "$letgo",
-          "date": "$(date -u +%Y-%m-%d)"
-        }
-        JSON
+      run: bash tools/write-build-info.sh "${{ inputs.dist }}"
--- a/.github/actions/build-wasm/action.yml
+++ b/.github/actions/build-wasm/action.yml
@@ -45,3 +45,22 @@ runs:
       env:
         XSOFY_WASM_INDEX: ${{ inputs.dist }}/index.html
       run: let-go tools/patch_wasm_coi.lg
+
+    # 4. Build provenance for the shell's debug-bar chip. The shell fetches
+    #    build-info.json; generating it here (the single bundle build) keeps the
+    #    deploy, the smoke gate, and PR previews consistent and stops the fetch
+    #    from 404-ing — the smoke gate counts any console error as a failure.
+    - name: Write build-info.json
+      shell: bash
+      run: |
+        letgo="unknown"
+        if [ -f .let-go-version ]; then
+          letgo="$(awk '!/^[[:space:]]*(#|$)/ {print $1; exit}' .let-go-version)"
+        fi
+        cat > "${{ inputs.dist }}/build-info.json" <<JSON
+        {
+          "xsofy": "$(git rev-parse HEAD)",
+          "let-go": "$letgo",
+          "date": "$(date -u +%Y-%m-%d)"
+        }
+        JSON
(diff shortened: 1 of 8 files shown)


mparrett:
Review finding — **P2: replay boot bypasses the browser HUD initialization path.**

`emit-startup` is called only from `new-game`, but `play-replay` constructs its world through `build-game`. The replay reduction also renders directly instead of going through `play/play-advance`, so the HUD title, quest, and stats remain placeholders throughout playback. On entering `game-loop`, stats finally update, but title and quest never do.

Please emit startup data from the common construction/replay path; if the HUD is expected to track replay playback, emit stats for those rendered replay turns as well.

Relevant code: https://github.com/nooga/xsofy/blob/8ef62cfa17896b2467e0170f02a6d9e358cccd5d/main.lg#L336-L369

mparrett:
Confirmed and fixed in `5f585ae`.

`emit-startup` moved into `assemble-world`, the construction path both `new-game`
and `build-game` (replay) funnel through, so the HUD title and quest now populate
for replay as well. And since you noted stats should track playback, `play-replay`
now emits stats for the initial world and each rendered replay turn, so the HUD
follows along instead of jumping to live values only on entering `game-loop`.

Verified the HUD still populates on a normal run after the move (title, quest,
depth, and hp all set); replay reaches the same `assemble-world` call, so it gets
the startup emit for free.

Heads-up: this is the base of the #114 → #115 stack, so I rebased both onto it
and force-pushed. Their diffs are unchanged (main.lg vs the shell file are
disjoint) — just new base SHAs.

mparrett:
Heads-up on why this branch moved (now `b2ad700`): two issues on the downstream
previews trace back to here, so I fixed them at the base.

- `build-info.json` was never generated by the bundle build, so the shell's
  fetch 404'd (a console error the preview smoke rejects). Added the generation
  to the `build-wasm` action.
- The `refit` BEL nudge could dismiss the title screen; guarded the title
  against a bare BEL in `title.lg`.

Both cascade to #114 and #115. The replay-HUD fix (`5f585ae`) is untouched.

mparrett:
## Review findings

- **P1 — landscape touch devices have no controls by default.** Touch controls are enabled only for portrait or `force-touch` (`tools/xsofy-shell.html:211–231`, `271–274`). A phone/tablet rotated to landscape loses the feature this PR introduces. Gate on coarse pointer/touch capability, using orientation only for layout.

- **P1 — this drops existing mobile resize coverage.** The base handled debounced `resize`, `orientationchange`, and `visualViewport`; #113 replaces that with one synchronous window listener (`tools/xsofy-shell.html:463–505`). #114 restores it, so #113 must not ship independently.

- **P2 — local browser verification is broken.** Manifest generation exists only in the GitHub action, while `make e2e` and `make browser-smoke` build through the Makefile’s `wasm` target, which does not create `build-info.json`. On the exact head, all four browser tests failed from that 404; after supplying the manifest, all four passed.

- **P2 — BEL filtering remains incomplete.** Resize BELs are filtered for title/help/codex, but confirmation reads still treat BEL as “no” (`main.lg:500–540`). Resizing silently cancels hazard, restart, and quit confirmations. Use the filtered read helper for these prompts too.

### Verdict

Request changes.

Validation: exact head `a9b1cd7`; 2,725 assertions and native smoke passed. Browser suite passed 4/4 after adding the missing manifest.

mparrett:

Issue #83, opened 2026-06-22
Browser: resize / orientation / visualViewport fit


mparrett:
**Layer:** browser shell (xsofy client shell once `-w-shell none` migration lands; today the default `let-go/pkg/rt/wasm/lg-shell-xterm.js`)
**Parent:** nooga/xsofy#84 (Player interaction ergonomics).

## Problem

The shell's fit path is a single naive listener:

```js
window.addEventListener('resize', () => fitAddon.fit());
```

No debounce, no `orientationchange`, no `visualViewport`. On phones this misses the cases that actually matter: the URL bar showing/hiding fires `visualViewport` resize (not always `window` resize), rotation needs an `orientationchange` settle, and an un-debounced `fit()` on every resize event thrashes during a drag/rotate and can land on an intermediate size. The viewport-meta + `dvh` work (#291) fixed the *height*; the *re-fit* trigger is still the weak link.

## Repro

1. Phone browser, portrait: scroll so the URL bar collapses → terminal doesn't re-fit to the reclaimed height until another event fires.
2. Rotate portrait→landscape → grid may settle on a stale column count.
3. Show/hide on-screen touch controls → no re-fit.

## Fix shape

Replace the single listener with a debounced fit driven by `resize`, `orientationchange`, and `visualViewport` resize/scroll, plus an explicit call after touch-control show/hide and font change. Post-settle debounce (the handoff's "reload-or-two to establish COI" aside is separate — this is purely the fit trigger). One `requestAnimationFrame`-coalesced `fit()` after the settle, then forward the resulting cols/rows to the host once.

## Acceptance

- Re-fits correctly across: rotate, URL-bar show/hide, font-size cycle, touch-control toggle.
- No fit thrash mid-drag (debounced/coalesced — one fit per settle).
- Cols/rows forwarded to the host match the rendered grid after settle (no off-by-one ghost column/row).
- Desktop browser resize still fits; native terminal unaffected.

## Dependencies

- Builds on #291 (viewport meta + `dvh`).
- Lands cleanest after / alongside the client-shell migration (#78, `-w-shell none`) so the fit logic lives in xsofy's own shell rather than patched into the let-go default.
- Pairs with the modal-input child (loop stays live to handle resize while menus are open).

mparrett:
Fixed: shell now debounces refit on resize/orientationchange/visualViewport (100ms, coalesces bursts) via `74360eb`, on top of #143. Verified against current `origin/main`.

PR #114, opened 2026-07-02
Debounced resize / orientation / visualViewport re-fit (#83)

mparrett:
# Debounced resize / orientation / visualViewport re-fit

Addresses #83. Stacked on #113 (the touch shell) — the re-fit function and the
explicit toggle/font-cycle calls this builds on land there; review this diff as
the top commit.

## Why

The shell re-fits off a single `window` `resize` listener. On a phone that
misses the cases that matter:

- The URL bar showing or hiding reclaims height but doesn't always fire a
  `window` resize — it fires a `visualViewport` resize. The grid never re-fits
  to the reclaimed rows.
- Rotation settles over a burst of events; a fit on each can land on an
  intermediate size.
- An un-debounced fit thrashes mid-gesture during a drag or rotate.

The viewport-meta and `dvh` work fixed the height; the re-fit *trigger* was
still the weak link.

## What

Route the ambient triggers (`resize`, `orientationchange`, and `visualViewport`
resize/scroll) through a trailing-edge debounce that coalesces into a single
`requestAnimationFrame` fit per settle. `refit()` is already gated on an actual
character-grid change, so no-op events (a `visualViewport` scroll that doesn't
change the row count) cost nothing. Discrete user actions (the font-size cycle
and the touch-control toggle) keep calling `refit()` directly; they're single
events and don't need settling.

## Validation

Browser, mobile viewport (390 wide):

- Grow height 700 → 980: grid re-fits 16 → 23 rows. Shrink to 500: → 11 rows.
- A burst of five rapid resizes coalesces to two fits (not five), and the final
  grid matches the final viewport — debounced, no thrash.
- `crossOriginIsolated` holds, terminal boots, no new console errors.

Desktop resize still fits (same `resize` path, now debounced); native is
unaffected — none of these events exist there.

### DIFF
--- a/tools/xsofy-shell.html
+++ b/tools/xsofy-shell.html
@@ -521,7 +521,34 @@
       }
     });
 
-    window.addEventListener("resize", refit);
+    // Ambient viewport changes — drag-resize, rotation, and the mobile URL bar
+    // showing/hiding — arrive as a burst of events over the gesture. Debounce
+    // to the trailing edge and coalesce the fit into one rAF so it runs once
+    // per settle, never mid-gesture or on an intermediate size. Discrete user
+    // actions (font cycle, touch-control toggle) call refit() directly below —
+    // they're single events and don't need settling.
+    let refitTimer = null, refitRaf = null;
+    function scheduleRefit() {
+      if (refitTimer) clearTimeout(refitTimer);
+      refitTimer = setTimeout(() => {
+        refitTimer = null;
+        if (refitRaf) cancelAnimationFrame(refitRaf);
+        refitRaf = requestAnimationFrame(() => { refitRaf = null; refit(); });
+      }, 120);
+    }
+
+    window.addEventListener("resize", scheduleRefit);
+    // Rotation settles over several resize events; the trailing-edge debounce
+    // waits them out, so no extra orientation-specific delay is needed.
+    window.addEventListener("orientationchange", scheduleRefit);
+    // The URL bar collapsing/expanding reclaims height without always firing a
+    // window resize; it does fire visualViewport resize (and scroll on some
+    // engines). refit() is grid-change-gated, so scroll spam costs nothing when
+    // the row count is unchanged.
+    if (window.visualViewport) {
+      window.visualViewport.addEventListener("resize", scheduleRefit);
+      window.visualViewport.addEventListener("scroll", scheduleRefit);
+    }
 
     // Shell-owned font sizing — replaces the fork's host-side _lgSetFontSize.
     // Safe before term.open (sets the option; the fit() is caught); the


mparrett:
## Corrected review follow-up

After tracing the shell history, the team is correct: the debounce and mobile viewport listeners predate this stack.

- `74360eb` originally added a 100 ms debounced `resize` / `orientationchange` / `visualViewport.resize` refit.
- The #113 shell replacement (`8ef62cf`) accidentally dropped that behavior, leaving only a direct `window.resize → refit` listener while adding the BEL resize nudge.
- #114 restores the prior behavior as a 120 ms + `requestAnimationFrame` debounce and additionally handles `visualViewport.scroll`.

The title-dismissal failure was real and reproducible, but my original attribution to #114 was incorrect. The underlying incompatibility was between the BEL nudge introduced in #113 and the existing title-screen “any key” handler; #114 merely exposed it more reliably through the restored triggers. The fix correctly landed on the #113 base (`b2ad700`) and #114 inherited it through its rebase.

The `build-info.json` 404 was also fixed in the shared base build action. The rebased #114 now passes deploy, pinned, and latest checks.

**Status: both findings resolved; no remaining findings on #114.**

mparrett:
Both fixed, and verified through `smoke.mjs` itself.

**Title dismissal.** The nudge is a lone BEL (`0x07`); the title's "any key" loop
counted it as a keypress. Guarded the title against a bare BEL in `title.lg` —
real keys still start the game. Matched BEL exactly rather than via `parse-key`,
since space also parses to `:unknown` and has to keep working. The fix lands on
the base branch (`mobile/bridge-shell`), which owns `refit` and the BEL nudge, so
this PR inherits it through the rebase.

**build-info.json 404.** The composite bundle build never wrote it. The deploy
runs that same `build-wasm` action, so the chip has been dead in production too,
not only in previews. Added the generation to `build-wasm` (the single source
for deploy, smoke, and preview), so the fetch 200s everywhere.

Ran the gate on the integrated stack: `#terminal` at 909ms, `Press any key` at
9406ms, zero console errors.

Rebased the stack onto the fixed base: this PR is now `2d96a74`, #115 `df6ff0b`.

mparrett:
Clarifying what this PR changes relative to `main`, since the framing as fresh resize/fit work overstates it.

`origin/main` already debounced resize handling: a 100ms `setTimeout` feeding `fit.fit()`, wired to `resize`, `orientationchange`, and `visualViewport` resize. #113's shell rewrite (landing the touch UI) rebuilt `refit()` to add the worker binding and, in the process, kept only the `window.resize` handler — it dropped the debounce, the `orientationchange` listener, and the `visualViewport` listeners.

So most of this PR is restoration of parity with `main` that #113's rewrite temporarily dropped: the debounce (now a 120ms trailing edge coalesced into a `requestAnimationFrame`), `orientationchange`, and `visualViewport` resize. The new piece is the `visualViewport` scroll listener, for the mobile URL bar showing and hiding.

Net state on the merged stack is correct (the debounce comes back here), so no code change is needed. Noting it so the history reads accurately and #113's intermediate regression is on record rather than looking like this PR invented resize handling from scratch.

mparrett:
## Review result

No new correctness findings. The debounce and `resize` / `orientationchange` / `visualViewport` wiring correctly restore the behavior lost in #113.

Validation on exact head `64b27de`:

- 2,725 assertions passed.
- Native smoke passed.
- A focused browser check confirmed the terminal grid grows and shrinks after portrait viewport height changes.

Verdict: approve this diff, but the stack remains blocked by #113’s findings. A committed resize regression test would be worthwhile.

mparrett:

Issue #84, opened 2026-06-22
Epic: Player interaction ergonomics


mparrett:
Umbrella for how it *feels* to play xsofy across the native terminal, desktop browser, and phone browser. The shared problem is interaction ergonomics: input semantics, modal menu flow, resize/fit behavior, readable browser display/fonts, smooth touch/keyboard response, and menus that still work on small displays — the difference between a roguelike that boots and one that's actually playable on each surface.

The work splits into three layers, tracked as grouped checklists below: each holds its filed children, proposed children (filed as slices firm up), and related context. The split is **by layer, not platform**: a shell change spans desktop + mobile, so platform would cross-cut the layers. A layer graduates to its own sub-epic issue only if its list outgrows this body.

## E1 · TUI / game loop

Input semantics, modal menu flow, replay-safe actions, resize-aware rendering.

**Filed**
- [ ] nooga/xsofy#82: non-blocking modal/menu input. Convert the remaining blocking `term/read-key` modal paths to nested `xsofy.ui/run-loop`s, with replay/determinism checks. Closes the rest of nooga/xsofy#56.
- [ ] nooga/xsofy#181: UI design first pass — menu, prompt, and screen consistency. A survey of every player-visible screen, prompt, and HUD panel, plus 14 findings on footer grammar, dismiss keys, key vocabulary, palette drift, and empty-state wording. A separate axis from the rest of this epic: how the UI *reads and looks* rather than how it *feels* to interact. Mostly E1, but findings 4 and 7 compare the terminal against the touch pane, so it reaches into E2.

**Proposed** (each gets its own repro, fix shape, and acceptance checks when filed)
- [ ] Small-display layout & text wrapping: sidebar/log/status collapse thresholds; long quest/log/help/menu wrapping; prompt fitting at narrow sizes.
- [ ] Resize-aware camera/rendering: recenter/invalidate the camera on resize, fit, and font-size change; no stale tiles, ghost rows, or off-screen player.

**Context**
- nooga/xsofy#56: Responsive UI event loop (composable, replay-safe input + cosmetic animation). Mostly landed on `main` (`xsofy.ui/run-loop`, title/death screens, core play loop, terminal docs); the remaining modal/nested-loop conversion is #82.

## E2 · Browser shell

xterm ownership, font metrics, viewport fitting, touch controls, interaction smoothness.

**Filed**
[...]

PR #115, opened 2026-07-02
Center the terminal grid (letterbox the fit remainder)

mparrett:
# Center the terminal grid (letterbox the fit remainder)

Stacked on #114 — it extends the same `refit()`. Part of the #84 shell
ergonomics work; review this diff as the top commit.

## Why

xterm renders a fixed-cell grid, and the FitAddon floors to a whole number of
columns and rows. A container is almost never an exact multiple of the cell
size, so up to ~one cell is left over on each axis — and xterm pins the grid to
the top-left, pooling that slack on the right and bottom edges. At most viewport
sizes the grid sits visibly hard against the top-left with a ragged gap on the
right and below.

## What

In `refit()`, split the remainder on each axis into equal padding on `#terminal`
so the grid sits centered. The padding is cleared before `fit()` (so it measures
the full box, not the previously-padded one) and re-applied after. `onReady` now
calls `refit()` once so the centering also lands on first paint, not only after
a resize.

`#terminal`'s border-box is fixed by flex (height) and the default column-flex stretch
(width), so the padding stays inside it and never shifts the bar below it. The
grid geometry is untouched — same columns, same rows, same cell metrics. Only
the offset changes, so xterm's selection and mouse mapping stay aligned: they
read the screen element's own rect, which moves with the padding.

## Validation

Browser, measured grid offset (px):

- Landscape 1440×760: was 0/19 right, 0/11 bottom → **9/10** left/right, **5/6**
  top/bottom. Grid stays inside `#terminal`; the bar below is unmoved.
- Portrait 390×780: **8/8** and **7/8**; the `55vh` terminal height is
  preserved, so the grid centers within it and the touch panel below is
  untouched.
- Re-splits on resize (e.g. 1607 wide → 9/9).

The ≤1px asymmetry is the odd-pixel remainder of an odd slack.

### DIFF
--- a/tools/xsofy-shell.html
+++ b/tools/xsofy-shell.html
@@ -480,7 +480,31 @@
     // Mirrors native SIGWINCH behavior; gated on grid change so per-pixel
     // resize events don't spam the worker.
     function refit() {
+      // Clear the letterbox padding before fitting so fit() measures the full
+      // box, not the previously-padded one (which would ratchet inward).
+      const host = term.element ? term.element.parentElement : null; // #terminal
+      if (host) { host.style.padding = "0px"; }
       try { fit.fit(); } catch (_) {}
+      // Letterbox the sub-cell fit remainder on both axes. The grid is a whole
+      // number of fixed-size cells, so up to ~one cell is left over per axis;
+      // xterm pins the grid top-left, pooling the slack on the right and bottom
+      // edges. Split each axis into equal padding so the grid sits centered —
+      // the geometry is unchanged, only where the slack lives. #terminal's
+      // border-box is fixed by flex (height) and stretch (width), so this
+      // padding stays inside it and never shifts the bar below.
+      const screen = host ? host.querySelector(".xterm-screen") : null;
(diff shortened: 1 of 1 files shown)


mparrett:
No independent findings in #115. Its letterboxing measured correctly in landscape and portrait without overflow or moving the shell, but it inherits #114’s title-screen regression.

mparrett:
The inherited title regression is fixed at the base (a BEL guard in `title.lg`),
so nothing changes in #115 itself — it just rebased onto the fixed stack and is
now `df6ff0b`. The letterboxing is unchanged. Smoke passes on the assembled
stack: `Press any key` renders, zero console errors.

mparrett:
Reviewed [PR #115](https://github.com/nooga/xsofy/pull/115): no new correctness findings. The centering logic avoids padding accumulation and the full native suite passes (2,725 assertions), as do native smoke and WASM build. The stack remains blocked by inherited #113 issues. A focused browser geometry check couldn’t run.

mparrett:

PR #119, opened 2026-07-02
Narrow-width top bar — in-grid HP/status when sidebar folds

mparrett:
Builds on #112 (the responsive layout collapse) and is based on that branch, so the diff here is just the top bar. If #112 lands first, this retargets to `main`.

## The problem

Below 80 columns, #112 folds the sidebar to give the map room. But the sidebar is the only in-grid home for HP and status effects. Once it's gone, that information survives only in the browser chrome HUD — so native and SSH terminals lose it entirely, and even in the browser it disappears if the HUD bridge never connects (embedded in a webview without the JS host, for instance). HP is not decoration; dropping it silently on a narrow terminal is a real gap.

## The change

When the sidebar folds, reclaim one row at the top of the game rect for HP and status. It is left-aligned and deliberately does not fill the width:

- a fixed-width HP bar, at the same width and colors the sidebar already uses;
- status-effect badges after it, each painted on its own effect color, so a green `poi` reads as poison at a glance;
- depth.

The trigger is width only — `show-topbar?` is `(not show-sidebar?)` — so the bar is mutually exclusive with the sidebar and wide layouts are untouched.

## Why it composes cleanly

- It reuses the existing HP-bar primitive and the status palette, so it matches the sidebar's look — no new rendering path.
- It renders in the terminal grid, so native TUI and WASM get it identically. The browser chrome HUD becomes redundant rather than load-bearing; removing that duplicate HP is a separate follow-up, sequenced after this so HP is never left without a home.
- It takes row 0 before the map/log/status split. The camera and letterbox both operate on the map rect, so neither is affected — the map just starts one row lower.

## Reviewing

Reviewable in a native terminal with no shell. Run at a narrow width (say 40×24 or 30×30) so the sidebar folds; the top bar then carries HP and depth. Status badges appear once the player has an active effect (poison, burning, and so on).

Badges use a three-character tag. A one-cell-per-effect variant for ultra-narrow rows can come later as an automatic overflow fallback.

### DIFF
--- a/xsofy/render.lg
+++ b/xsofy/render.lg
@@ -473,6 +473,58 @@
             (let [next-row (render-creature-entry r @row c)]
               (reset! row (or next-row (+ @row 2)))))))))))
 
+;; --- Top Bar (narrow-width HP/status; replaces the folded sidebar) ---
+;; One row, left-aligned (it does NOT fill the width): a fixed-width HP bar
+;; using the same width + colors as the sidebar's, then status-effect badges,
+;; then depth. Renders in the terminal grid, so native TUI and WASM get it
+;; identically — the browser chrome HUD is no longer HP's only home.
+
+;; Status badge: a 3-char tag ("poi", "bur"), painted on the effect's own bg
+;; color so the color carries the signal. A one-cell-per-effect variant for
+;; ultra-narrow rows can come later as an automatic overflow fallback.
+(defn- topbar-badge-text [st]
+  (let [nm (:name st)]
+    (subs nm 0 (min 3 (count nm)))))
+
+(defn render-topbar [world r]
+  (when r
+    (ui/clear-rect r)
+    (when-let [player (get-in world [:entities :player])]
+      (let [hp     (or (get-in player [:body :hp]) 0)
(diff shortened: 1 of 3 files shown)


mparrett:
## Review findings

- **P2 — the new top bar violates the 10-row map floor at 67–79×14.** The log threshold still assumes only log + status consume rows, but the new top bar consumes another row afterward (`xsofy/ui.lg:143–164`). At `67×14`, layout produces a 9-row map; `66×14` and `80×14` produce 10. Base the threshold on `body-h` or include `topbar-h`.

- **P2 — an invisible status badge can suppress depth.** When a badge does not fit, the cursor still advances (`xsofy/render.lg:498–513`). At 22–23 columns, depth fits with no status, but activating one status draws neither the badge nor depth. Stop advancing when a badge is clipped or reserve depth before badges.

### Verdict

Request changes. The PR correctly closes #112’s primary HP/status gap, but it also inherits #112’s camera-reset blocker.

Validation: exact head `19215ea`; 2,725 assertions and runtime smoke passed. No top-bar/layout tests were added.

mparrett:

PR #120, opened 2026-07-02
Gate dev tooling behind ?mode=; keep player options visible

mparrett:
Builds on #113 (the touch shell + HUD bridge, where this chrome lives) and is
based on that branch, so the diff here is just the mode change. It doesn't
depend on #114 or #115 — they don't touch this part of the shell.

## The problem

The info bar was all-or-nothing. One `TOGGLES` tile flipped between a play view
and a debug view, and the debug view lumped player preferences (font size,
hold-to-repeat) together with dev-only diagnostics (build string, perf counters,
force-touch). A default player saw either none of it or all of it — including a
build-provenance string and per-turn render timings that mean nothing to them.

## The change

A three-tier ladder selected by a `?mode=` URL param, mapped to stacked root
classes on `#xsofy-shell`:

- **play** (default, `mode-play`) — player options only: font size, hold-to-repeat.
- **debug** (`?mode=debug`, adds `mode-debug`) — adds player-visible diagnostics:
  build provenance, perf counters. Also reachable at runtime by tapping the tile.
- **dev** (`?mode=dev`, adds `mode-dev`) — adds dev-only tooling: force-touch,
  which reveals the phone touch UI on desktop or tablet-landscape for testing.
  URL-only; the tile never cycles into it, so dev tooling stays out of a
  player's reach.

The classes are additive (`play ⊂ debug ⊂ dev`), so each tier keeps everything
the lower tiers show. Chip and stat visibility is gated on `.debug-only` /
`.dev-only` under those classes.

Two cleanups ride along: the tile is renamed `TOGGLES → OPTIONS`, since it now
fronts real player preferences; and the old `data-page` chip-paging is retired,
superseded by the tier classes. An old `xsofy/shell-state` value of `toggles`
migrates to `debug`.

## Scope

Shell-only — no `.lg` change. A real cheats panel would belong in the dev tier,
but that's deferred: `cheats.lg` is a graft-only overlay, not part of the
shipping build, so wiring it in would pull in game-side code that isn't present.
This adds the tier seam and reserves the dev slot.

## Reviewing

Load the shell three ways and check the info bar (the chip row is bottom-right):

- default — font + hold-to-repeat + the OPTIONS tile; no build string, no perf
  counters, no force-touch.
- `?mode=debug` — adds the build string and perf counters; still no force-touch.
- `?mode=dev` — adds force-touch on top.

Tapping the OPTIONS tile cycles play ↔ debug (a player can reach the
diagnostics); `dev` is only reachable via `?mode=dev`.

### DIFF
--- a/tools/xsofy-shell.html
+++ b/tools/xsofy-shell.html
@@ -1003,8 +1003,10 @@
 
     function applyShellState() {
       // mode-play stays as the permanent base class on the div; only the higher
-      // tiers toggle. dev is independent of the play↔debug cycle.
-      shellRoot.classList.toggle("mode-debug", currentShellState === "debug");
+      // tiers toggle. dev is a strict superset of debug — ?mode=dev pins mode-debug
+      // on regardless of the play↔debug cycle, so dev never shows less than debug
+      // (its force-touch chrome can't appear without the debug diagnostics under it).
+      shellRoot.classList.toggle("mode-debug", currentShellState === "debug" || devMode);
       shellRoot.classList.toggle("mode-dev", devMode);
       if (cycleDotsEl) {
         cycleDotsEl.textContent = SHELL_STATES.map((s) => s === currentShellState ? "●" : "○").join(" ");
--- a/tools/xsofy-shell.html
+++ b/tools/xsofy-shell.html
@@ -113,10 +113,9 @@
      (d/hp/turn-counter); the chip cluster sits bottom-right at thumb
      reach. Each text element only competes with the small fixed-width
      thing in its own row, so neither truncates the other.
-     Chip-row width is bounded by paging — see the .shell-cycle tile +
-     per-page data-chip-page tags below. Only one virtual chip group is
-     visible at a time, so the row stays single-line in debug even with
-     8+ debug chips defined.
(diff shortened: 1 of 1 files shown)


mparrett:
Reviewed at `c9a6147`.

One finding:

- **`?mode=dev` leaves an active PLAY/OPTIONS control whose transition is a visual no-op.** `applyShellState()` pins `mode-debug` whenever `devMode` is true (`tools/xsofy-shell.html:984`), but the pointer handler still cycles `currentShellState`, updates the dots/label, and persists the new state (`:998-1004`). Verified in Chromium: dev mode starts with `○ ● PLAY →`; clicking it changes the control to `● ○ OPTIONS →`, while the build/perf diagnostics and force-touch control remain visible and the play title/stats never return. The control therefore reports a transition that did not happen, and it silently changes the next no-parameter session via localStorage. Disable/hide the cycle in dev mode, or make its dev-mode transition affect the visible tier coherently.

Validation on the exact head: 279 tests / 2,725 assertions pass, native smoke passes, and the WASM build succeeds. The local build still logs the inherited missing `build-info.json` 404; that is not introduced by this PR.

mparrett:

PR #116, opened 2026-07-02
[preview only · do not merge] mobile stack integration (#111–#115, #119–#123) + #124 title fix

mparrett:
**This is not a merge candidate.** It's the mobile-stack PRs assembled onto
`main` so the whole thing can be played in one build — no single PR contains all
of it. Review happens on the individual PRs below; this one exists only to carry
a `deploy-preview` build.

📄 **Review writeup** — design rationale, per-PR breakdown, before/after by terminal size (rendered): https://raw.githack.com/nooga/xsofy/1b1f0976443be1ef6363fba4ac41c6209bbbd7db/docs/mobile-stack-review.html

🚀 View preview at
https://nooga.github.io/xsofy/pr-preview/pr-116/

## What it aggregates

- #112 — responsive camera + narrow-terminal rendering
- #119 — narrow-width top bar (in-grid HP/status when the sidebar folds); stacks on #112
- #111 — native engine benchmark + `xsofy.perf` core
- #113 — mobile touch shell + browser HUD bridge
- #114 — debounced resize / orientation / visualViewport re-fit (closes #83)
- #115 — center the terminal grid (letterbox the fit remainder)
- #120 — gate dev tooling behind `?mode=`; keep player options visible; stacks on #113
- #121 — drop redundant HP/depth from the browser chrome (now that the top bar carries them); stacks on #120
- #122 — reflow the title and death screens on resize / font-change (they captured the terminal size once and never re-read it); sibling off #113
- #123 — OPTIONS panel: player options behind a ⚙/← icon tile (pinned top-right, on its own); the perf stats and the option chips (font, hold-to-repeat, force-touch) share one horizontal row beneath the tile; turn/perf/dims gated to `?mode=debug`/`dev` and self-hidden until they carry a real value (no dead `t·`/`@t·`/`·×·`); build chip tap-revealed; fixed-tall info bar; stacks on #121

Also folded into this preview build (not part of the mobile stack, tracked as its own PR): #124 — keep the title-screen decorative runes off the title/prompt/seed text; they were blanking the seed line as they pulsed.

Two lines feed in: #112 with #119 stacked on it forms the trunk (they share
`ui.lg`/`render.lg`, so #119 stacks rather than merges). The shell line
(`#113 → #114 → #115`, with `#120 → #121 → #123` and the #122 screen-resize fix
also branching off #113) and the independent bench #111 merge onto that trunk.
The build here is the composite deploy build, the same one a release ships.

## Viewing

[...]

PR #121, opened 2026-07-02
Drop redundant HP/depth from the browser chrome

mparrett:
Builds on #120 (dev-gating) and is based on that branch — it's the tip of the
shell line. **Depends on #119** (the narrow-width top bar): the removal is only
safe where the top bar is present, so this should merge after #119. See the
sequencing note below.

## The problem

The browser info bar duplicates HP and depth that the game already renders in
the grid. That duplication is a leftover: before the top bar, the chrome was the
only place HP showed once the sidebar folded on a narrow viewport. With #119, HP
and depth have an in-grid home at every width — the sidebar when wide, the top
bar when the sidebar folds — so the chrome copies are redundant.

## The change

Remove the `hp` / `max-hp` / `depth` spans from the chrome info bar and the JS
that updates them. Title and quest stay — the grid has no home for those.

The ui-bridge still emits `hp` / `max-hp` / `depth` (the same stats event also
carries the debug turn counter), so the bridge contract is unchanged; the values
are no longer displayed. Shell-only, no `.lg` change.

## Sequencing — depends on #119

The top bar (#119) is on the #112 game line; this chrome lives on the #113 shell
line. They're parallel, and the removal is only safe when both are in the build:
a build of the shell line alone, without the top bar, would show no HP on a
narrow viewport.

Verified on an integration build carrying both #119 and this cut: HP renders in
the grid at narrow (top bar, `HP 30/30 … D1`) and wide (sidebar, `HP 30/30`),
the chrome HP element is gone, and title/quest remain. Review this via the #116
integration preview, which has the top bar; this slice's own build (the shell
line, no top bar) is not a faithful mobile preview on its own.

### DIFF
--- a/tools/xsofy-shell.html
+++ b/tools/xsofy-shell.html
@@ -376,8 +376,11 @@
     <span class="quest play-only" id="xs-quest"></span>
     <span class="build debug-only" id="xs-build" title="Build provenance — written by local-scripts/deploy.sh"></span>
     <span class="stats">
-      <span class="play-only"><span class="k">d</span> <span id="xs-depth">·</span></span>
-      <span class="play-only"><span class="k">hp</span> <span id="xs-hp">·</span>/<span id="xs-max">·</span></span>
+      <!-- HP and depth used to live here (play-only), but the in-grid top bar
+           (sidebar when wide, top bar when narrow) now carries both, so the
+           browser chrome no longer duplicates them. title + quest stay — the
+           grid has no home for those. The bridge still emits hp/depth; they're
+           just not displayed here. -->
       <span class="debug-only"><span class="k">t</span><span id="xs-turn">·</span></span>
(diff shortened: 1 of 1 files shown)


mparrett:
Reviewed at `66534de`.

No independent correctness findings in this diff. The removed DOM references and event updates are internally consistent:

- On the exact #121 head at a wide viewport, HP/depth remain in the in-grid sidebar, the browser chrome no longer duplicates them, and startup/stats events run without a new exception.
- On the current #116 assembled stack at 390×844, #119's narrow top bar renders `HP 30/30` and `D1` while the browser chrome remains free of those duplicate values.

The sequencing constraint is real: the current #121 ancestry does not contain #119, so this slice by itself removes the only narrow-width HP/depth display. It is safe only when the final merge also contains #119, as the PR description states. It also inherits the active-but-no-op PLAY/OPTIONS transition under `?mode=dev` reported on #120.

Validation on the exact #121 head: 279 tests / 2,725 assertions pass, native smoke passes, and the WASM build succeeds.

mparrett:

PR #122, opened 2026-07-02
Reflow title and death screens on resize / font change

mparrett:
Based on #113 (it builds on the title/death screens' input handling, which #113
guards). Independent of #114/#115/#120.

## The problem

The animated wait screens — title and death — capture `[tw th]` and scatter
their runes once, before the poll-based run-loop, and never re-read the terminal
size. So a resize or font-size change while sitting on either screen leaves it
frozen at the size it opened at: runes clustered for the old grid, text
off-center for the new one. `play.lg` already re-reads `term/size` every tick and
repaints; these screens didn't. The browser shell's font-size control makes it
easy to hit — cycle the font on the title and it doesn't reflow.

## The change

Thread `:size` and `:runes` through the loop state, and on the first tick after a
size change, re-scatter the runes and re-clear for the new dimensions. The
render reads size and runes from state instead of the captured values.

The resize path already wakes the loop — the shell nudges a blocked read-key with
a lone BEL, which the screens ignore (and still do). The size delta detected on
the next tick is what drives the re-layout.

The descending screen is a brief auto-advancing transition, so a resize there is
fleeting; it's left as-is.

## Verifying

Load the browser build, sit on the title screen, and cycle the font size: the
runes re-scatter and the title re-centers for the new grid. Before this, the
screen stayed frozen at the launch size.

### DIFF
--- a/main.lg
+++ b/main.lg
@@ -454,8 +454,7 @@
 
          :death
          (let [[tw th] (term/size)
-               runes (render/death-runes tw th)  ;; scatter ONCE — stable across frames
-               _ (sfx/clear-screen tw th)  ;; clear ONCE, not per-frame — was the flicker source
+               _ (sfx/clear-screen tw th)  ;; initial clear; the loop re-clears + re-scatters on resize
                ;; Dev repro capture of this organic run's replay code. Trigger is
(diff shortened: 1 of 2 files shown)


mparrett:

PR #123, opened 2026-07-03
OPTIONS panel — tiered controls, self-revealing diagnostics

mparrett:
Stacked on #121 (`shell/chrome-stats-cut`). **Draft** — pending native-TUI and WASM validation before review.

## What

Reworks the browser shell's info bar into a single **OPTIONS ↔ PLAY** tile that reveals a tiered control set, instead of always-on chrome. Three tiers, additive (`play ⊂ debug ⊂ dev`):

- **play** — title + quest only. The tile reads `OPTIONS →`.
- **debug** (tap the tile) — build provenance + perf counters + font/hold-to-repeat controls. Tile flips to `PLAY →`.
- **dev** (`?mode=dev`) — adds force-touch. URL-gated, so dev tooling never appears from a tap.

Perf cells are **self-revealing**: each stays hidden until its datum is emitted, so the bar shows live numbers rather than dead placeholders.

## Why these choices

- **Tier gating is CSS-class-driven** (`mode-play`/`mode-debug`/`mode-dev` on the root), so a higher tier keeps everything the lower ones show. `?mode=dev` pins `mode-debug` on regardless of the tile's play↔debug cycle — otherwise the dev-only force-touch chip could show while the debug diagnostics under it were toggled off, i.e. dev showing *less* than debug.
- **The info bar is pinned to a fixed height** (60px landscape / 74px portrait). The two tiers carry different content (play: title+quest; debug: perf row + a two-line build chip + taller font/repeat chips), so a content-driven height twitched on every toggle. Flooring above each orientation's taller tier keeps the toggle from resizing the bar.
- **The build chip wraps to two labelled rows** (`xsofy <sha> ·<date>` / `lg <version>`) rather than truncating with an ellipsis — the let-go version is often long, and both rows now name their repo so the shas aren't ambiguous.

## Not included

Title/death-screen resize reflow is **#122**'s scope, not this PR.

### DIFF
--- a/tools/xsofy-shell.html
+++ b/tools/xsofy-shell.html
@@ -107,15 +107,14 @@
   /* Two-row info bar. Single-row crammed title/quest/stats/chips into
      ~375px and lost the right half of both proper-noun strings on every
      phone. Grid layout:
-       row 1:  [ title ]   [ stats ]
-       row 2:  [ quest ]   [ chips ]
-     Stats sit next to the title because they're both gameplay info
-     (d/hp/turn-counter); the chip cluster sits bottom-right at thumb
-     reach. Each text element only competes with the small fixed-width
-     thing in its own row, so neither truncates the other.
-     The chip row holds the player options (font, repeat) plus the OPTIONS
-     tile always, and the dev-only force-touch chip under ?mode=dev — a small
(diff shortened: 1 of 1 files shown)


mparrett:

PR #124, opened 2026-07-03
fix(title): keep decorative runes off the title/prompt/seed text

mparrett:
## The bug

On the title screen, the decorative pulsing runes draw *over* the "seed NNNN" line (and can nick "press any key" and the title), obscuring characters.

It looks like a draw-order problem but isn't — the text already draws after the runes each frame. The real cause: `screenfx/scatter-runes` places ~40 runes at random cells with no collision check, so some land on the text rows, and once a rune fades below the brightness cutoff it **redraws its own cell as a blank space every frame**. That blank lands on top of the just-drawn text, so the seed line flickers and drops characters as runes pulse over it.

## The fix

Carve a "moat" around the text and skip any rune inside it, so runes never draw or blank over the title/prompt/seed. The box covers all three text rows (title at `cy`, prompt at `cy+2`, seed at `cy+4`) plus a one-cell ring, sized to the widest line. `sline`/`ssx` are hoisted into the `let` so the moat computation and the seed render share one definition.

## Verifying

Load the title screen at a seed (e.g. `?seed=350950991`) and let the intro settle: the title, "press any key", and "seed …" all render on a clean band with no runes touching them.

### DIFF
--- a/xsofy/title.lg
+++ b/xsofy/title.lg
@@ -107,23 +107,40 @@
         cy (quot th 2)
         tx (max 1 (- cx (quot (count title) 2)))
         sx (max 1 (- cx (quot (count subtitle) 2)))
-        title-fade (min 1.0 (/ (float frame) 20.0))]
-    ;; draw pulsing runes — each rune redraws only its own cell
+        title-fade (min 1.0 (/ (float frame) 20.0))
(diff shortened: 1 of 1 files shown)


mparrett:

PR #137, opened 2026-07-05
feat(mobile): touch shell redesign — portrait + landscape

mparrett:
**Scope:** the complete mobile touch shell against the new client shell (#154's branch). The controls foundation folded in from #155 (this branch already carried its commit), the landscape split this PR started as, the relayout folded from #156, and the portrait + landscape redesign on top — reviewing the intermediate states separately stopped making sense, and this diff is the end state (the superseded OPTIONS tile nets out entirely). Lands by squashing into `shell/app-mount`, growing #154. Nearly everything is in `tools/xsofy-shell.html`; the rest is a `version` field in `tools/write-build-info.sh`, a stale-builder guard in the `Makefile`, and a responsive e2e spec.

## Portrait

- **Keyboard-slice model.** The terminal caps at `55dvh`; the buttons pack into the fixed slice below it (`grid-auto-rows: minmax(0, 1fr)`). The slice never grows to fit content, both panes are exactly 5 rows, and every key fills its row, so pane swaps don't resize anything, the game area stays constant, and the grid cannot overflow into a scrollbar. The fixed-slice invariant replaced a long seesaw of height-tuning; it's the part to preserve in future edits.
- **`a·z` pane is a real QWERTY keyboard** (20-unit fine grid, 10 keys per row, home/bottom-row stagger, ⌫ in the bottom stagger slot, an esc·space·?·⏎ function row). The full letter set is load-bearing: menus label items `a..z` up to `max-inventory 26` (`entities.lg:248`, `main.lg:41`, `render.lg:852`), so anything smaller can't select a full drop menu. Keys are ~37px on a 390px phone, below the 44px guideline; OS keyboards make the same tradeoff.
- **PLAY pane**: D-pad right with a strong amber fill (vi-key annotations kept), menu letters `a b c d e` down column 2 (`a/d/e` double as USE/DROP/EQUIP), FIRE/GET + stairs + RUNES in column 3, `esc` and `? HELP` anchored top-left.
- **Top-bar switcher, two segments (`config | a·z`)** with PLAY as the unlit base state: each segment toggles its layer, tap again to return. Dropping the third segment freed the bar for title + quest on a 390px phone. The `⚙ config` pane holds the player/dev chips that used to crowd the top bar.

iphone13-portrait

## Landscape

- **Rail layout.** The ~30% landscape rail fits 4 key columns, not 6 (the old grid clipped the whole right D-pad column behind a scrollbar). Panes now carry a landscape-specific tile list, re-rendered on rotate: rotating into landscape previously kept the portrait grid because only the terminal refit. The 4x6 arrangement: pad right-middle, action column down the left, y/n/wait beneath the pad (a corner-anchored-pad alternative lost the on-device comparison). It cuts `b`/`c`, the only tiles with no game binding.
- **`a·z` opens a full-width bottom sheet** instead of squeezing into the rail (10 keys across ~250px is ~25px slivers). The rail keeps the PLAY grid while the sheet is up, so movement stays under the thumb during menus; a second `a·z` tap dismisses.
- **Status strip.** A 22px band under the terminal takes the info-bar content in landscape: title + quest left, diagnostic chips right. The rail info row keeps just the switcher.

iphone13-landscape

## Build provenance (#44)

`write-build-info.sh` (shared by `make wasm` and the build-wasm action, so tagged Pages builds get it with no workflow change) now emits a `version` field: nearest `v*` tag via `git describe` (exactly the tag on a release build, tag-distance-sha on dev builds). The shell surfaces it twice: the full provenance chip moved from the debug info bar to the bottom of the config pane, ungated (bug reporters shouldn't need a URL param), and the release part (`xsofy v0.0.2`) stands in for the title until the game announces itself.

Mode tiers are now URL-only. The retired OPTIONS cycle tile persisted play/debug in localStorage; with the tile gone, any phone that had ever tapped it was pinned in debug (title invisibly hidden in play) with no UI to leave. `?mode=` is the single source of truth; a future config-pane toggle brings its own persistence.

## Review focus

- `renderTouch()` and the `layoutFor()`/`placeStats()` pair: landscape-rail mode selection, the sheet render path, and the title/quest/stats reparenting between the info bar and the status strip.
- The font-size tap clears the viewport, but only when cols/rows actually changed: xterm reflows the old frame into the new grid (a garbled flash until the game repaints), while a same-dims tap sends no resize nudge, so an unconditional clear would sit blank until the next keypress.
- Ergonomics were tuned on a real phone; emulator widths floor at ~500px, so key-size judgments came from device screenshots.

## Screenshots

Being added — portrait PLAY and a·z, the landscape rail, the bottom sheet, and the config pane.

### Portrait (390×844)

| PLAY | a·z | config |
| --- | --- | --- |
|  image  |  image  |  image  |

### Landscape (844×390)

| rail (PLAY) | a·z bottom sheet |
| --- | --- |
|  image  |  image  |

### Compact + desktop

| desktop | desktop force-touch | desktop responsive |
| --- | --- | --- |
|  image  |  image  |  image |

## Open questions

- The touch/desktop visibility matrix (what shows under force-touch on desktop, narrow desktop windows, etc.) gets a coherent pass in a follow-up.
- ⌫ sends DEL (`0x7f`); it taps cleanly but no menu has confirmed consuming it yet.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,11 @@ wasm: ## Build the WASM web app into $(DIST): client-owned shell + COI bridge
 	# and the js/url-param host seam (#339) that carries ?seed=/?replay= into the
 	# worker natively.
 	$(LG) -w $(DIST) -w-shell none main.lg
+	# Guard: a stale PATH `lg` (pre let-go#378) emits the old #terminal host
+	# mount; the injected shell then dereferences a missing #app and the bundle
+	# never boots. Fail here instead of shipping it (see the .let-go-version
+	# NOTE — CI checks out the pin, local builds use whatever `lg` is on PATH).
+	@grep -q 'id="app"' $(DIST)/index.html || { echo 'wasm: $(DIST)/index.html has no #app mount — stale lg builder? (.let-go-version)'; exit 1; }
 	XSOFY_WASM_INDEX=$(DIST)/index.html $(LG) tools/inject_shell.lg
 	# COI bounded-retry still patches the core glue: it drives the crossOriginIsolated
 	# lifecycle, which the shell contract (window.LetGoHost) and a static host (e.g.
--- a/tests/e2e/responsive.spec.js
+++ b/tests/e2e/responsive.spec.js
@@ -1,0 +1,106 @@
+const { test, expect } = require('@playwright/test');
+const { spawn } = require('child_process');
+const path = require('path');
+const http = require('http');
+
+// Responsive coverage for the touch shell: phone portrait plus three landscape
+// heights down to iPhone-SE class. Each viewport boots the bundle with touch
+// emulation, cycles the layer switcher (config, a·z, back to play), and
+// asserts the shell never overflows its box — the compact-landscape clipping
+// regression rode exactly this gap (rail scrollHeight > viewport at 667x375).
+
+function serve() {
+  return spawn(process.execPath, [path.join(__dirname, 'static-server.js')], {
+    env: { ...process.env, COI: '1', PORT: '8123',
+           DIST: process.env.DIST || path.join(__dirname, 'dist') },
+    stdio: 'pipe',
+  });
+}
+
+async function waitForServer(port = 8123, timeout = 5000) {
+  const start = Date.now();
+  while (Date.now() - start < timeout) {
+    try {
+      await new Promise((resolve, reject) => {
+        http.get(`http://localhost:${port}/`, (res) => { res.destroy(); resolve(); }).on('error', reject);
+      });
+      return;
+    } catch (e) { await new Promise(r => setTimeout(r, 100)); }
+  }
+  throw new Error(`Server did not start within ${timeout}ms`);
+}
+
+function stopServer(srv) {
+  return new Promise((resolve) => { srv.on('exit', () => resolve()); srv.kill(); });
+}
+
+function watchConsole(page) {
+  const errors = [];
+  page.on('console', (m) => { if (m.type() === 'error') errors.push(m.text()); });
+  page.on('pageerror', (e) => errors.push(String(e)));
+  return errors;
+}
+
+// No scroll anywhere: the document doesn't scroll on either axis, and the
+// shell (the rail, in landscape) fits its own box. 1px of tolerance for
+// subpixel rounding on fractional dvh/percentage splits.
+async function expectNoOverflow(page, label) {
+  const m = await page.evaluate(() => {
+    const doc = document.scrollingElement;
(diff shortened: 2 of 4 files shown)


nooga (approved):

mparrett:
Independent pre-PR review findings from the local `mobile/touch-relayout-fold` branch:

1. **P1 — compact landscape rail overflow.** At 667×375 the shell measured `scrollHeight 384` inside a 375px viewport; at 568×320 the touch grid measured 193×384 inside 169×320, clipping the bottom row and right column. The short-landscape override retained `flex: 0 0 auto` and the 44px key-width floor.  
   **Current PR status:** addressed by `8f6f16a` (`fix(mobile): compact landscape rail fits short phones; device safe areas`).

2. **P2 — incomplete safe-area handling.** With `viewport-fit=cover`, portrait controls lacked a bottom inset and the landscape rail/status sheet lacked left/right inset handling, risking controls beneath the home indicator or unsafe screen edges on notched devices.  
   **Current PR status:** addressed by `8f6f16a`.

3. **P2 — pinned local builder was not enforced.** I reproduced a successful-looking `make wasm` with a stale `lg` that emitted `#terminal`; the injected shell expected `#app`, threw `Cannot read properties of null (reading 'style')`, and never mounted the terminal. CI used the correct pin, but local/release builds needed a guard.  
   **Current PR status:** addressed by `cb102cd` (`build: fail wasm build when the #app host mount is missing`).

The original review also recommended responsive regression coverage. The PR now includes `2ef4f71`, covering touch pane switching and overflow across four phone viewports.

Original validation: exact pinned WASM build succeeded; runtime smoke passed; 284 tests / 2,738 assertions passed with zero failures or errors; 390×844 and 844×390 portrait/landscape pane flows were console-clean.

nnunley (approved):

nnunley (approved):

mparrett:
@nooga Just checking in since some refinements were added since your last approval. Let me know if it still looks good and I'll get this merged.

mparrett:
Before merging, for the record — both approvals technically predate the branch tip:

- @nnunley approved at `fe832c4`; the only change since is the rebase onto main after #154 merged. The tip patch is byte-identical.
- @nooga approved on Jul 6, before the redesign commits that followed; I checked in on the 16th and don't want to block on a re-review given the final state has a fresh approval.

CI is green on the rebased head, and I re-verified the live deploy boots and plays end to end. Squash-merging (the intermediate layouts never stood alone); leaving `mobile/landscape` in place since #150 targets it.

PR #140, opened 2026-07-05
GL frontend experiment: hardware-rendered xsofy + level-2 fall into first person

nnunley:
Experiment branch (draft): the whole game rendered through a minimal OpenGL platform layer, with entering depth 2 tipping the playfield into the floor of a 3D world and continuing in first person.

Runs on the let-go `glplat` branch build (nooga/let-go#392; async parity in nooga/let-go#391):

    XSOFY_SEED=42 <let-go>/lg main_gl.lg

## What's in the branch
- `main_gl.lg` on the unified poll-based event loop (xsofy.ui, #56): glplat input source + clock source, pure step fold, render subscriber — same tick-vs-input contract as the terminal, so cosmetic animation never touches the deterministic (seed, action-log) stream.
- Presentation state machine: :title (pulsing rune particles, any key) → :top-down → :falling (2.5s camera descent + ortho→perspective morph on real stairs descent) → :fp (grid-step, turn-based, billboarded entities).
- `xsofy/gl/`: mat4, grid adapter (floor quads + extruded walls tinted from the lit tile palette with per-face shading), bridge (FOV/lighting/stains/gases), fall morph, fpview, vfx overlays (projectiles/bolts/flashes), HUD (status line, HP bar, log tail; spell-rune glyphs render 2-cells-wide with 1-cell advance so inscriptions overlap into composite sigils, matching the native-terminal look).
- Atlas pipeline in lg: `tools/bake_atlas.lg` + `assets/gl/fonts.edn` bake Fairfax HD (the web shell's single face) at its native 12x24 grid with binarized alpha; `ui/chan-source` added for push-based producers now that let-go has poll!.
- Env hooks for headless verification: XSOFY_GL_AUTOEXIT, XSOFY_GL_SHOT[_AT] (frame → PNG), XSOFY_GL_FORCE_FALL, XSOFY_GL_SKIP_TITLE, XSOFY_GL_DUMP.

Suite green throughout (416 tests, 2988 pass). Known gaps: no browser/WebGL backend yet, screenfx tint flash stub, no smooth FP camera lerp, no FOV culling of submitted geometry.


mparrett:
We rebased this branch onto current `main` for a downstream experiment and profiled the top-down path along the way. The results are in #180, based on this branch's tip (`cc43ffb`) rather than a rebase, so it merges with a button and nothing of yours moves.

The headline: culling cells that are neither in FOV nor remembered takes the top-down frame from ~400ms to ~18ms at seed 42, and the captured frame is byte-identical before and after. Also included are frame timing behind `XSOFY_GL_PERF` and two input fixes — every keypress was moving two tiles, and the arrow keys were dead entirely.

Independent of that PR: the test suite hangs on this branch's current base. It reaches `decode . encode = identity on (seed, actions)` in `replay-test` and then allocates without bound, 12.4 GB before we killed it. Clean `cc43ffb` reproduces it with none of our commits applied, and current `main` runs the same suite in 60 seconds with 444 tests and 3,066 assertions green, so it appears to be fixed upstream already. That may be worth a rebase on its own.

nooga:
let's undraft, we are doing it 🚀

mparrett:
@nnunley #140 had fallen 46 commits behind `main`, so I rebased it for you to take or leave. It's your PR and your branch, so I won't push to `gl-frontend`; how it lands is your call.

The branch is [`mparrett/xsofy:experiment/gl-frontend-v2`](https://github.com/mparrett/xsofy/tree/experiment/gl-frontend-v2): your 16 commits, then #180's 4, cherry-picked onto `main` at `0c38395`, plus two new commits at the end. Your commits keep you as author; the committer field shows me because the cherry-pick ran here.

Three ways it could go, in the order I'd suggest:

1. **Force-update #140 from it.** One push, and #180 closes as folded in:

```
git fetch https://github.com/mparrett/xsofy experiment/gl-frontend-v2
git push --force-with-lease=gl-frontend:cc43ffb origin FETCH_HEAD:gl-frontend
```

2. **Redo the rebase yourself** and use that branch as the reference answer. Only two files conflicted:
   - `xsofy/test/run.lg`: `main` switched to `run-all-tests` for let-go 1.13's `clojure.test` port. I kept `main`'s form and appended your five test namespaces.
   - `xsofy/title.lg`: `main` wrapped each frame of `show-descending-screen` in `sfx/with-frame` (#174). I kept `main`'s body and grafted in only your `XSOFY_GL_FRONTEND` guard. Taking your side would have dropped both `with-frame` calls. The guard counts characters because `os/getenv` returns `""` for an unset variable, not `nil`.
3. **A replacement PR from that branch**, if you'd rather hand it off.

## Verified on let-go `main`

Built let-go `main` (`36d57e36`) twice, with `-tags glplat` and with `-tags glplat_ebiten`, and ran the branch on each (macOS, arm64):

| Check | GLFW | Ebitengine |
| --- | --- | --- |
| Suite | 475 tests, 0 failures | same |
| Boot to gameplay, clean exit | yes | yes |
| Title, fall, first person | yes | renders; first person mis-sorts (no depth buffer) |
| Culled top-down frame byte-identical to pre-cull | yes | yes |
| Cells / cpu per frame, seed 42 | 2,370 → 40, ~365 → ~17 ms | 2,370 → 40, ~382 → ~17 ms |
| One move per keypress, Shift+K runs, arrows move and repeat | yes | yes |
| Glyphs upright | yes | yes |

The `replay-test` hang nooga saw in #180's `latest` CI leg on 2026-09-20 came from the old base; the suite finishes on the new one.

## The two new commits

- `2d57261` moves `parse-glplat-event` out of `main_gl.lg` into a pure `xsofy.gl.events` namespace, since `main_gl.lg` runs `-main` on load and nothing could require it from a test. `events-test` covers the #180 input fixes, which nooga noted had no test. Putting the double-move back fails it.
- `c535ab3` ignores bare modifier keys. Both backends send `key:lshift` and the like, which dispatched as an unbound key and cleared `:running`, so pressing Shift mid-run stopped the run. Ebitengine also auto-repeats modifiers, so holding Shift there stopped it about 20 times a second.

## Two questions

- **How does a player run it?** The PR body points at the `glplat` branch build from let-go#392, which is closed. Today it takes let-go `main` built with `-tags glplat` (cgo, GLFW) or `-tags glplat_ebiten`, then `lg main_gl.lg`; released `lg` binaries register no backend. Do you want a short "running the GL frontend" section, and should it name one backend as the default?
- **First person on Ebitengine.** C4 in let-go's backend contract makes painter's order the portable guarantee, and first person currently leans on the GLFW depth buffer. Would you ship first person as GLFW-only for now, or hold it for a back-to-front sort?

PR #144, opened 2026-07-08
feat(console): read-only dev command console (Layer 1)

mparrett:
## What

Reshapes this PR from an eval REPL into a **read-only command console** — the first of three layers.

A Quake-style dev console (backtick, dev-gated) as a game-side `:screen`. Input is parsed with the real reader, but the head is dispatched against a fixed whitelist — there is no `eval`, so a typed form can only run a curated, side-effect-free command. The world goes in and comes back unchanged, so the console can't fold the seed, grow the `:action-log`, or taint a run. The safety is structural, not policed.

## Commands

All pure world-map reads, with no coupling beyond the UI and world-as-data:

- `(seed)` — seed-input + current rolling seed
- `(player)` — hp / position / depth
- `(location)` — position / depth / turn
- `(help)` — the command list

Both `(help)` and a bare `help` resolve; an unknown or non-whitelisted form is reported, never run.

## Why the split

The original single PR bundled the console UI, read-only reads, subsystem-integrating commands, and an arbitrary-eval path. Splitting by concern:

- **Layer 1 (this PR)** — the console + pure reads. No external dependency.
- **Layer 2 (stacked, #153)** — `(replay)` and `(bench)`, which pull in `xsofy.replay`, `xsofy.perf`, and the browser bridge.
- **Layer 3** — the arbitrary-eval REPL with mutations recorded as commits in the game-state DAG (@nnunley's review below). Its reference implementation is preserved on the `console/dev-repl-cut1` branch.

## Validation

Full lg suite green (2756 assertions). Read-only verified: running every command against a live world leaves it byte-identical.

### DIFF
--- a/main.lg
+++ b/main.lg
@@ -17,6 +17,7 @@
             [xsofy.screenfx :as sfx]
             [xsofy.debug :as dbg]
             [xsofy.dispatch :as dispatch]
+            [xsofy.console :as console]
             [xsofy.seed :as seed]
             [xsofy.replay :as replay]
             [xsofy.ui-bridge :as bridge]))
@@ -556,6 +557,29 @@
                  (render/render-full w)
                  (recur w w (term/size)))
                w)))
+
+         :console
+         ;; Quake-style dev REPL. The game is already on screen (the play loop
+         ;; just drew it and toggling the console doesn't change the world), so
+         ;; we do NOT repaint it here — the panel just draws over the top rows.
+         ;; The only full repaint in the open/close cycle is on close, to erase
+         ;; the panel. Runs on its own poll loop with a RAW-key source (line
+         ;; editing needs literal chars, not parsed game actions). Returns the
+         ;; world unchanged — the console is read-only (Cut 0). See xsofy.console.
+         (let [size (term/size)
+               result (ui/run-loop
+                        ;; frame-ms 30 matches the play loop's input cadence (it
+                        ;; floors to 30ms under wasm anyway) so typing feels the
+                        ;; same as the game. Cheap now that render-console is
+                        ;; dirty-checked — a faster loop won't repaint while idle.
+                        {:sources  [(ui/input-source ui/raw-key->event) (ui/clock-source 0)]
+                         :state    (console/init-state world size)
+                         :step     console/handle-key
+                         :render   console/render-console
+                         :frame-ms 30})
+               w (dissoc result :screen)]
+           (render/render-full w)
+           (recur w w (term/size)))
 
          ;; default: game screen — the poll-based turn loop (input +
          ;; clock-driven auto-actions / ambient animation); see run-play-loop.
--- a/xsofy/console.lg
+++ b/xsofy/console.lg
@@ -1,0 +1,270 @@
+(ns xsofy.console
+  "Quake-style dev console: a read-only command surface over the running game.
+
+   It is a game-side :screen (see xsofy.main game-loop), driven through the same
+   poll-based event loop (xsofy.ui/run-loop) as every other screen, so it works
+   identically in the native TUI and in the browser (wasm) with no shell change
+   and no build flag. Backtick opens it; it is dev-gated (see xsofy.input/dev?).
+
+   READ-ONLY by construction (Cut 0). Input is parsed with the real reader but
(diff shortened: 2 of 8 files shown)


mparrett:
@nnunley Curious to get your comments here, especially with regard to the replay/dag part.

mparrett:
Review findings:

1. **P1 — replay dumps are invalid after console use.** `main.lg` appends a human-readable note directly to the URL-safe Base64 replay code. The emitted/printed value then contains spaces and brackets, so passing it to `replay/decode` fails. Keep `code` raw and emit the truncated/console-used note separately (for example, a second native stderr line and a separate field in the web event).

2. **P2 — the taint guard is bypassable from the REPL.** `xsofy.console` evaluates forms in the same namespace that holds the live-world `repl` atom. A form can mutate `(:world @repl)` directly, including `aset` on terrain, without calling `w` or `w!`; `:tainted?` remains false, no `:console-taint` marker is written, and later replay data can silently diverge. Since `repl` is directly reachable despite being private, use a more conservative taint policy or an evaluator-state design that does not expose the live world.

nnunley:
So... Thoughts:
1. One thing you can do is make mutation only happen through a protected namespace. You should be able to create a shadowed set of the general APIs that only allows your own specially controlled version of the xsofy namespace be modified.
2. The dag was supposed to have two kinds of commits/changes.
- player inputs
- special full snapshots.
You should be able to make anything that mutates state... or more cruelly, views state, add a command or special commit to the DAG. Which in turn could be acted upon later. Anything that actually mutates state should still go through the command stream for the game. Having true god mode should require a command line option to make most repl commands traceless.

Tainting could just be including the special player repl input marker for the command. You already have a strong hash of priors, so you should be able to tell when tampering has happened (incorrectly).

mparrett:
Reshaped this along the lines of your DAG feedback, split into layers so each concern lands on its own:

- **This PR (Layer 1)** is now a read-only command console — no `eval`. Input is parsed with the real reader but dispatched against a fixed whitelist, so a typed form can only run a curated read (`(seed)`, `(player)`, `(location)`, `(help)`). The world comes back unchanged, so there's no taint path at all. That sidesteps the determinism question for the common "just let me see state" case.
- **Layer 2** (stacked) adds `(replay)` and `(bench)`.
- **Layer 3** is the arbitrary-eval REPL, and that's where your points land: mutations recorded as commits in the game-state DAG (the `checkpoint-hash` / external-injection case `xsofy.dag` already anticipates), the protected mutation surface, and traceless god-mode behind a flag. I've kept the original eval REPL on `console/dev-repl-cut1` as the reference to build it from.

So the taint-and-refuse hack is gone from the shipping path; the "record the intervention as a commit" design you described becomes Layer 3 rather than something bolted onto this one.

mparrett:
Thanks — this is the model I want Layer 3 to follow, and it lines up with where this stack is headed.

This PR (Layer 1) sidesteps the whole question by having no `eval`: the reader parses input, but the head only dispatches against a fixed read-only whitelist, so a typed form can't reach a mutating API at all. Safety is structural, not policed. That keeps the console surface landable on its own while the mutation story gets built properly on top.

On your three points, for the eval layer:

**Protected namespace.** Agreed that's the right shape for the write surface — the read whitelist here is the read-only half of the same idea, and the mutating half wants your shadowed-API approach so only a controlled `xsofy` surface is reachable, never the general runtime.

**Everything through the command stream.** This is already the plan: mutations land as commits in the game-state DAG (`xsofy.dag`), and traceless god-mode is a deliberate opt-in, not the default — a CLI flag, as you said. The "or more cruelly, views as commits too" angle is a good one I hadn't drawn all the way out; recording a read as a DAG event makes the repl session itself replayable, which is worth it for repro even when nothing mutates.

**Tainting + tamper detection.** Same page — a repl-marked command carries the taint, and the DAG's strong prior hash is what makes tampering detectable when the marker is missing or the chain doesn't reconcile. So the marker gives provenance and the hash gives integrity.

The reference implementation for that layer is parked on `console/dev-repl-cut1`, and the `xsofy.console` / `xsofy.dag` docstrings point at the external-injection / checkpoint-hash case so the intent survives the split. I'll build Layer 3 against this comment. Keeping this PR read-only so it can merge on its own; the mutation surface comes next.

nnunley (approved):

PR #145, opened 2026-07-08
perf(render): elide redundant SGR escapes in the map render

mparrett:
## What

Adds a small render-only "pen" to the map render that skips re-emitting a `set-fg`/`set-bg` whose color already matches the last one emitted. `move-cursor` and `write` are untouched, and iteration order is unchanged, so the output is the same — same glyphs, same colors, fewer escape sequences.

## Why

`term`'s `set-fg`/`set-bg` write an SGR escape on every call, and the map loop calls them for every cell unconditionally. SGR attributes persist across cursor moves, so any cell whose color equals the previous one's is re-sending bytes for no visual change. On native that's wasted output; on the WASM build each call is a crossing into the JS host and xterm.js, so the waste is larger there.

## How it works

- **Eliding across cursor moves is safe.** SGR state is sticky: moving the cursor doesn't reset the pen color, so skipping a redundant `set-fg` is correct as long as nothing reset the terminal's color in between.
- **Reset discipline.** The pen mirrors the terminal's current color, valid only while every styled write goes through it. Panels, vfx overlays, and the  full-clear emit raw and leave it stale, so each contiguous penned batch resets the pen before its first cell.

## Where it helps

Elision fires where same-color cells are already adjacent in emit order, so it does most of its work on the full-screen repaints — floor change, camera pan, resize — which walk the map row-major and are the biggest single render spikes. The per-move FOV-delta path iterates in set order and elides little; that's left as-is on purpose. Sorting it into runs would raise the elision rate slightly, but `fire`/`water`/`lava` tiles draw their color from the global `rand` stream, so reordering iteration would change which cell consumes which draw. Harmless in itself (that stream is unseeded, so the shimmer already varies run to run) but a needless behavior change for ~1% fewer escapes, so base iteration order stands.

## Scope

Modest on its own — most frames are per-move FOV-delta renders, where there's little to elide. Its value is cutting the full-repaint spikes, and serving as the foundation for a frame-to-frame appearance diff (a follow-up PR stacked on this one), which is where the bulk of the render savings come from.

## Safety

Rendering is a pure side effect — it doesn't touch the seed, action log, or world state — so this cannot affect determinism or replay. Static terrain and glyphs render identically before and after (verified by comparing the displayed cells, colors included, over a fixed input sequence); animated tiles already vary run to run via the unseeded `rand` stream, unaffected either way. Existing test suite passes unchanged.

### DIFF
--- a/xsofy/render.lg
+++ b/xsofy/render.lg
@@ -67,6 +67,40 @@
 ;; the main game thread).
 (def last-camera (atom nil))
 
+;; --- SGR pen (render-perf: lossless set-fg/set-bg elision) ---
+;; let-go's `term` is a raw-emit primitive — every set-fg/set-bg writes an SGR
+;; escape immediately, so the map loop re-sends identical fg/bg for every cell
+;; even across same-color runs (walls, memory, blanked cells). SGR attributes
+;; persist across cursor moves, so re-emitting an unchanged color is pure waste —
+;; bytes on native, and a Go→JS→xterm.js crossing per call on WASM.
+;;
+;; The pen records the last fg/bg we emitted and skips a set-fg/set-bg whose
+;; color already matches. move-cursor and write are untouched, so this is
+;; lossless: identical glyphs, identical colors, just fewer escapes.
+;;
+;; Correctness: the pen mirrors the terminal's current fg/bg, valid only while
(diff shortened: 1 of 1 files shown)


mparrett:
Reviews complete:

nnunley (approved):

Issue #147, opened 2026-07-08
Epic: Render emit performance


mparrett:
**TL;DR:** a lane for cutting xsofy's terminal render cost — fewer escape sequences per frame, no change to what's drawn. A three-PR stack up, more scoped behind it. The payoff concentrates on the WASM/web build, where every emit crosses into xterm.js.

## What this tracks

Render-side emit efficiency in `render.lg`: trimming redundant terminal output without changing the rendered result. Distinct from the generation/native perf work (`perf/native-xxh3`, `perf/wander-lights-cache`), which targets map generation and the native build.

An independent lane off `main`: no dependency on the mobile shell epic (#128), upstream-mergeable on its own, and it flows back to every build once landed. The payoff is largest where mobile runs. On the WASM path each skipped emit is a Go→JS→xterm.js crossing avoided, on the most constrained target, so treat it as a mobile-readiness enabler judged on the phone-feel harness rather than native byte counts.

## Slices

- [x] **sgr-elision** (#145): a render-only pen that skips a redundant `set-fg`/`set-bg`; `move-cursor`/`write` and iteration order are unchanged, so it's lossless. Does most of its work on the full-screen repaints (floor change, camera pan, resize). Off `main`.
- [x] **appearance-diff** (#146): skip map cells unchanged since the last frame, since static lights mean most visible cells repeat frame to frame. ~35% fewer `set-fg`/`set-bg` and ~21% fewer bytes over a movement burst. Stacks on #145. Carries the golden tests (see below).
- [x] **synchronized output** (#149): wrap the frames that repaint in succession (`render-full`, `render-dirty`, and the vfx animation per frame) in DEC mode 2026 so each presents atomically, for anti-tearing most visible on xterm.js. Stacks on #146 (additive: keeps the emit reduction, adds atomic presentation). One-shot modal screens stay on a plain flush, since nothing streams after them to tear. Mode 2026 is a boolean (never nested); no feature detection (unsupported terminals ignore it); `sync-output?` is a source opt-out. A let-go `term` begin/end-sync primitive could factor out the escapes later if other TUIs want them, but xsofy owns the frame boundaries, so it lives here. Merged 2026-07-23.
- [ ] **back-buffer render model** (larger, scoped): a paint-to-buffer plus diff-on-flush layer that folds elision, the appearance diff, and the panel/entity/vfx paths into one place, making buffer coherence structural instead of the targeted invalidations #146 carries. Those invalidations are the argument for it.

[...]

PR #148, opened 2026-07-08
tool: lighttest — interactive lighting + render-perf testbed

mparrett:
## What

A standalone dev tool, `tools/lighttest.lg`, in the same spirit as `tools/mapgen.lg`: generate a dungeon with `xsofy.terrain`, drop a movable light, and render the lit result live. No game loop, monsters, items, or turns; it reuses only the pure `terrain`/`lighting`/`fov` functions plus terminal I/O.

lighttest-clean-half

Run it:

```
lg tools/lighttest.lg                # seed 1
lg tools/lighttest.lg '{:seed 42}'   # inline EDN config
```

## Why

It isolates the lighting model so you can see how falloff, radius, color mixing, and wall occlusion actually read, without the rest of the game in the frame. It also became the measurement vehicle for the render-perf lane (#147): the same testbed carries a live render A/B and the knobs used to characterize each change.

## What it does

- **Steer the light** with the keyboard (arrows / hjkl / yubn / numpad), or hand it to a screensaver demo. `p` cycles the demo path: off, Lissajous, edge-bounce, and map-scroll — the last scrolls the whole map under a fixed light, coherent full-screen motion used as the synchronized-output tearing stressor.
- **Lighting knobs:** radius, color cycle, wall occlusion (raw radial vs FOV shadows), terrain lights on/off, multiple lights.
- **Render-perf probe:** a three-way render A/B (`e` cycles delta, sgr-off, naive — sgr-off keeps the frame diff but re-emits SGR per cell, isolating the elision win alone) with a `drawn`/`sgr` readout, a color-depth posterize knob (`d`), and a synchronized-output toggle (`s`, DEC mode 2026). Every varying HUD field is fixed-width so the readout doesn't jitter while the light moves. `?` opens a full key reference.
- **Frame model:** a poll loop (`key-pending?` + `sleep`) that blocks when idle and animates when the demo is on; a terminal resize triggers a clean full repaint.

## Scope

Dev-only — it's not wired into the game or the build, a `tools/` entry you run directly. This was opened as a draft when probing the render-perf lane. Has mostly settled, so open for review now.

### DIFF
--- a/tools/lighttest.lg
+++ b/tools/lighttest.lg
@@ -1,0 +1,448 @@
+(ns lighttest
+  "Interactive lighting testbed — a dungeon and a movable light, nothing else.
+
+   Generates a level with xsofy.terrain, drops a light source you steer with
+   the keyboard (or let it fly on its own in demo mode), and renders the lit
+   result live so you can *see* how xsofy.lighting behaves — falloff, radius,
+   color mixing, and the gap between the raw radial light field and what
+   wall-occlusion (FOV) actually reveals. No monsters, combat, items, turns, or
+   dispatch: this reuses only the pure terrain/lighting/fov functions plus
+   terminal I/O.
+
+   Frame model: a poll loop (term/key-pending? + sleep). With demo off it blocks
+   on read-key (idle-cheap); with demo on it advances the light each frame and
+   drains input non-blockingly, so every knob stays live while it animates. A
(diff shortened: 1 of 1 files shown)


mparrett:
Nit: Occlusion should be on by default, toggle to disable

nnunley (approved):

PR #150, opened 2026-07-09
feat(shell): replay link in the config pane (#126)

### LINKED ISSUE #126: Dev: wire onEmit in the shell so ?dump-replay= surfaces the replay code on web
Follow-up split from #99 (closed). The player-facing bug there is fixed; this is the remaining dev-tooling gap it surfaced.

## Problem

#104 shipped `?dump-replay=` on web: on death the dev capture fires `(js/emit "xsofy/replay-dump" {:code ...})` instead of writing to the shared terminal writer. But the deployed `xsofy/tools/xsofy-shell.html` only wires `onReady` / `onOutput` / `sendInput` / `setSize` — it never registers `onEmit`. So on web the emit fires and nothing listens: the replay code is never surfaced (no console, no display).

Native capture works (`XSOFY_DUMP_REPLAY=1 <lg> main.lg 2>/tmp/replay.txt`); browser capture does not, as shipped.

## Fix shape

Register an `onEmit` handler in the shell that logs the replay-dump event:

```js
window.LetGoHost.onEmit?.((name, data) => {
  if (name === "xsofy/replay-dump") console.log("replay-code:", data.code);
});
```

Scope is the shell only — the game side (`js/emit`) already works. Dev-only affordance, so `console.log` is enough; no on-screen surface (that was the #99 bug).

## Acceptance

- With `?dump-replay=` set on the web build, a run that dies logs `replay-code: <code>` to the browser console.
- The code round-trips: pasting it into `?replay=` reproduces the run.
- No on-screen output on the death screen (the #99 regression stays fixed).

## Context

- Parent: #99 (closed).
- Web dump plumbing: #104.

mparrett:
Closes #126. Reframed from the earlier help-screen shape — #137's config pane is the natural surface now, so this stacks on #137.

## The gap (#126)

`?dump-replay=` fires `(js/emit "xsofy/replay-dump" {:code …})` on death, but nothing web-side listened: the code never surfaced on the primary (mobile) test surface. Native capture (`XSOFY_DUMP_REPLAY` → stderr) already works and is unchanged.

## This change

- Game: every web run captures live — a bridge emit every 10th turn (and on turn 1, so the chip is current from a run's first action), plus the death capture firing unconditionally on web for the exact final state. Death isn't special: `encode-run-safe` works on any live world. Native stays opt-in (`XSOFY_DUMP_REPLAY` / `?dump-replay=`).
- Shell: the emit becomes a chip in the ⚙ config pane — a `?replay=` link that reproduces the run in one tap (long-press for the OS copy/share sheet), carrying the session params (mode etc.; seed dropped — the code brings its own). Emits already arrive as window CustomEvents, so there's no new host plumbing. Unavailable codes (parameterized-action runs) keep the chip hidden rather than rendering a dead link, and the chip resets on every new run so it can never link a stale one. Not tier-gated: sharing/reproducing a run is a player feature.

## vs the earlier shape

The previous draft rendered a replay-code page inside the game's help screen. The config pane is the better surface: no game-render involvement, tap-to-reproduce instead of transcribing a code by hand, and the same coverage on the mobile test surface. The help-screen implementation is preserved on a local archive branch if the on-demand mid-run capture turns out to be wanted too.

## Verification

- Full loop in browser emulation against the built bundle: turn-1 emit fires, the chip reveals with the real code, and following the link boots the replay (HUD renders, dev chrome preserved). Chip stays hidden for unavailable codes and in the play tier. Native suite: 284 tests, 0 failures.
- Known sharp edge (pre-existing, out of scope here): an invalid `?replay=` code is a fatal decode error ("nth index out of bounds") — filed as #159.

_Config pane with the chip:

image

https://xsofy.quest/pr-preview/pr-150/?replay=AQAAAAAilCqbAwthdXRvZXhwbG9yZQVyaWdodARsZWZ0AAAAAwAUAQQCEA

### DIFF
--- a/main.lg
+++ b/main.lg
@@ -459,18 +459,19 @@
          :death
          (let [[tw th] (term/size)
                _ (sfx/clear-screen tw th)  ;; initial clear; the loop re-clears + re-scatters on resize
-               ;; Dev repro capture of this organic run's replay code. Trigger is
-               ;; per-platform: ?dump-replay= URL param on web (js/url-param),
-               ;; XSOFY_DUMP_REPLAY env var natively — and any dev build
-               ;; (?mode=dev), whose shell renders the emit as a replay link in
-               ;; the config pane. Gated so players never see it.
+               ;; Capture this organic run's replay code at its final state. On
+               ;; web this always fires: the shell's config pane renders the emit
+               ;; as a player-facing ?replay= link (the live per-10-turn capture
+               ;; in ui-bridge can lag a few turns; death re-emits exactly).
+               ;; Natively it stays opt-in via XSOFY_DUMP_REPLAY (or the
+               ;; ?dump-replay= param for parity).
                ;; Sink differs because *err* aliases *out* in WASM (one host writer
                ;; for both — let-go rendermain), so writing *err* there paints the
                ;; game screen. On web route the dump off-terminal via a js/emit host
                ;; event; natively keep stderr (redirect 2>file to capture past the TUI).
                _ (when-let [trig (or (js/url-param "dump-replay")
                                      (os/getenv "XSOFY_DUMP_REPLAY")
-                                     (when (= (js/url-param "mode") "dev") "dev"))]
+                                     (when (= (os/os-name) "js") "web"))]
                    (when (> (count trig) 0)
                      (let [code (replay/encode-run-safe world)
                            payload (or code "(unavailable: run has parameterized actions)")]
--- a/main.lg
+++ b/main.lg
@@ -461,13 +461,16 @@
                _ (sfx/clear-screen tw th)  ;; initial clear; the loop re-clears + re-scatters on resize
                ;; Dev repro capture of this organic run's replay code. Trigger is
                ;; per-platform: ?dump-replay= URL param on web (js/url-param),
-               ;; XSOFY_DUMP_REPLAY env var natively. Gated so players never see it.
+               ;; XSOFY_DUMP_REPLAY env var natively — and any dev build
+               ;; (?mode=dev), whose shell renders the emit as a replay link in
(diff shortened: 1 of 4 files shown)


mparrett:
@nnunley this is marked draft because I haven't had enough time to think hard about this. I'm not sure if the approach in this PR is the best way.

nnunley:
As a thought: you can include the current replay link in the browser window using our js bridge.

mparrett:
> As a thought: you can include the current replay link in the browser window using our js bridge.

Yeah, good idea. Probably should. I was thinking along the same lines some days ago, but feeling noncommittal about the UI shell as it existed. I thought at the very least, console.log it. I'll think a little more - if not in this PR, I'd land it in/after the one that introduces web UI.

mparrett:
Consider using xterm web-links addon / OSC 8:

https://xtermjs.org/docs/guides/link-handling/

nnunley:
So... Probably actually need a flash warning until you get your partial redraws code in.

mparrett:
> So... Probably actually need a flash warning until you get your partial redraws code in.

This is ready to go: https://github.com/nooga/xsofy/pull/151

mparrett:
151 was merged, this is now unblocked.

mparrett:
> Probably actually need a flash warning until you get your partial redraws code in.

@nnunley I want to make sure I didn't miss your intent here. I assumed you were referring to the strobing replay, which was fixed by the (now merged) redraw improvements. Let me know if there was anything else. I'm taking the OSC-8 terminal link as a follow-on, since it'll require a little more thought about where (and when) we want that link to appear in the terminal. I also need to catch myself up on the state of any relevant issues. This PR, for the web replay link, is good to go.

nnunley (approved):

PR #151, opened 2026-07-10
fix(render): incremental redraw during replay playback

### LINKED ISSUE #110: Web replay playback flickers (full-screen redraw each frame)
Playing back a run via `?replay=<code>` on the web build works — but the terminal does a full redraw each frame, so it flickers noticeably during playback.

Minor: playback is correct, it's purely visual. The replay path likely clears + repaints the whole screen each frame; a diff/partial redraw (or double-buffer) on playback would remove the flicker.

Context: first end-to-end confirmation that `?replay=` playback works on the web build (v0.0.2 candidate).

mparrett:
## What

`play-replay` now redraws incrementally during playback — a dirty FOV-delta render per turn — instead of full-repainting every frame. Full repaint stays only for a floor change, where the whole viewport turns over anyway.

Independent of the render-perf stack (#145/#146/#149): the change is in `play-replay`, not `render.lg`, and `render-dirty` is already the live path on `main`.

## Why

`render-full` blanks the entire terminal before it repaints — it paints every row with the background colour, then draws the panels on top. The live game loop only pays that cost when it has to (floor change or resize); every normal turn goes through `render-dirty`, which touches only the cells that changed and never clears.

`play-replay` never adopted that split — it called `render-full` on every action. So each replayed turn cleared the screen and repainted it, and on the web build that blank-then-repaint reads as a full-screen flicker across the whole playback (#110).

## How

Mirror `play/play-advance`: compare `(:depth w')` against the pre-action depth, `render-dirty w' w` when they match, `render-full` on a floor change. `render-dirty` takes the previous world as its diff base — the reduce already carries it, so threading it in is the whole change.

I left out the live loop's resize branch on purpose. Replay has no interactive resize mid-playback; a stray resize would leave one stale frame until the next turn redraws, which isn't worth polling `term/size` every frame for. Easy to add back if that turns out to matter.

## Safety

Render-only. The seed chain and action log are untouched, so a replayed run stays bit-identical — same reconstruction, only fewer escapes drawing it. Verified in the browser: the replayed run animates cell-by-cell with the map, HUD, and log staying put frame to frame, no full-screen blank between turns. Existing suite passes.

Closes #110

### DIFF
--- a/main.lg
+++ b/main.lg
@@ -387,7 +387,13 @@
     (sfx/pause! 400)
     (reduce (fn [w action]
               (let [w' (dispatch/dispatch w action)]
-                (render/render-full w')
+                ;; Mirror the live loop (play/play-advance): an incremental
+                ;; dirty render per turn, full repaint only on a floor change.
+                ;; A full repaint every frame clears the whole screen first,
+                ;; which reads as a blank-then-redraw flicker on playback (#110).
+                (if (= (:depth w') (:depth w))
+                  (render/render-dirty w' w)
+                  (render/render-full w'))
                 ;; Keep the HUD stats tracking the replay as it plays out.
                 (bridge/emit-stats w')
                 (sfx/pause! 90)


mparrett:
Confirmed fixed via this replay [pr-151 AQAAAAAA...](https://xsofy.quest/pr-preview/pr-151/?replay=AQAAAAAABnkyCAVyaWdodARkb3duBGxlZnQCdXAKZG93bi1yaWdodAh1cC1yaWdodAlkb3duLWxlZnQHdXAtbGVmdAAAAGAAAwECAgIDAQABAQEEAQUBBgEHAQABAQECAQMBAAEBAQADAQICAgMBAAEBAQQBBQEGAQcBAAEBAQIBAwEAAQEBAAMBAgICAwEAAQEBBAEFAQYBBwEAAQEBAgEDAQABAQEAAwECAgIDAQABAQEEAQUBBgEHAQABAQECAQMBAAEBAQADAQICAgMBAAEBAQQBBQEGAQcBAAEBAQIBAwEAAQEBAAMBAgICAwEAAQEBBAEFAQYBBwEAAQEBAgEDAQABAQE)

To see the problem, try same against 149 (if still open):

Confirmed fixed via this replay [pr-149 AQAAAAAA...](https://xsofy.quest/pr-preview/pr-149/?replay=AQAAAAAABnkyCAVyaWdodARkb3duBGxlZnQCdXAKZG93bi1yaWdodAh1cC1yaWdodAlkb3duLWxlZnQHdXAtbGVmdAAAAGAAAwECAgIDAQABAQEEAQUBBgEHAQABAQECAQMBAAEBAQADAQICAgMBAAEBAQQBBQEGAQcBAAEBAQIBAwEAAQEBAAMBAgICAwEAAQEBBAEFAQYBBwEAAQEBAgEDAQABAQEAAwECAgIDAQABAQEEAQUBBgEHAQABAQECAQMBAAEBAQADAQICAgMBAAEBAQQBBQEGAQcBAAEBAQIBAwEAAQEBAAMBAgICAwEAAQEBBAEFAQYBBwEAAQEBAgEDAQABAQE)

mparrett:
Note on merge order. 151, then 150. (this PR fixes the flicker that will be more visible once 150 lands)

See: https://github.com/nooga/xsofy/pull/150
cc: @nnunley

nnunley (approved):

PR #136, opened 2026-07-05
fit: debounced re-fit + letterbox the grid

mparrett:
Part of the mobile stack (umbrella: #128). Base: `mobile/shell`.

Keeps the grid fitted as the browser viewport changes: a debounced re-fit on resize, orientation, and visualViewport, plus a letterbox of the leftover space so the terminal grid stays centered.

## Non-obvious

- Web-specific, and debounced so a drag-resize or rotate doesn't thrash the layout.
- The letterbox centers the grid in whatever space is left after fitting whole cells, rather than leaving it pinned to a corner.

## Verify

Rotate the phone or resize the browser: the grid re-fits (debounced) and stays centered, with the remainder letterboxed.

PR #154, opened 2026-07-11
feat(shell): bind the #app host mount (let-go #378)

mparrett:
## What

Binds the client shell to the `#app` host-frame mount from let-go #378 (renamed from `#terminal`), and bumps `.let-go-version` to the commit that carries it. Terminal-only shipping shell — no feature work.

## Why now

The mobile-shell stack (#135→#136→#137) is written against `#app` and pins this same let-go commit, so today the mount migration is buried inside a touch-UI feature PR. That's what makes #135 conflict with #143 (the portrait shell landed on `main` still binding `#terminal`).

Pulling the migration out as its own base fixes that: the mobile stack rebases onto a `#app` `main` and drops its now-redundant `.let-go-version` bump, leaving the touch UI as the only thing it carries. It also clears the `#terminal`/`#app` drift for every other shell change in flight, not just mobile.

## Scope

Five lines in `tools/xsofy-shell.html` — the CSS cap selector, the `getElementById` lookup, and a host comment — plus the `.let-go-version` pin. The portrait 55vh cap is unchanged apart from the selector name.

## The pin is a SHA, not a tag

No release carries #378 yet (v1.11.1 predates it), so `.let-go-version` points at `cf195a6`. The file's note flags re-pinning to a release tag once let-go ships one. The deploy resolves its let-go build ref straight from that file, so the single pin bump flips the built `lg`; a stale prebuilt `lg` would silently emit the old `#terminal` frame, which is why the pin and the shell move together in one commit.

## Verification

Built deploy-faithfully (`go install let-go@cf195a6`, the same fetch the Pages workflow runs): the core emits `#app`, the shell binds it, xterm opens and fills the container 0.978×0.987 (the sub-cell remainder is absorbed by the letterbox), no console errors.

---

**Status (2026-07-13):** the stack around this PR consolidated — #152 folded in here (closed; both its commits were already ancestors), #135 landed into this branch, and the whole mobile shell is now one PR (#137) stacked directly on it. Landing order: #137 squashes into shell/app-mount, then this PR carries everything into main.


nnunley:
So, this works, but I'm noticing that the death screen doesn't have the replay link? Is that because your replay change haven't landed yet?

image

nnunley (approved):
lgtm

mparrett:
Thanks for the review!

> So, this works, but I'm noticing that the death screen doesn't have the replay link? Is that because your replay change haven't landed yet?

Yeah, I ran into a design question to discuss, so it'll land separately.

TL;DR - Exposing the replay link the web shell works, and we should do it. But I also want a clean route for TUI: OSC-8 is probably the nicest way. We can even make that work in xterm with the web-links addon. But like you said, the web shell itself is a nice place for such a thing. My open PR #150 currently does it that way.

Also bumping the related #110 which is an easy and nice enhancement to replay mode. :)

PR #155, opened 2026-07-12
feat(mobile): touch controls, OPTIONS panel, diagnostics + theming

mparrett:
Second slice of the rebuilt mobile stack, stacked on #135 (the blank scaffold).
Fills the scaffold with the actual mobile UI — the half held back so the
structure could land reviewable first. Purely additive on #135's containers and
mode-class contract; the two plumbing one-shots #135 left inert grow their
runtime toggles here.

**What's here**
- Theming (`--bar-*` vars, the info-bar two-row grid, button/tile chrome) + the
  JetBrains mono terminal face.
- Info bar: title + quest (left-ellipsized to keep the item name), the OPTIONS
  ⚙↔← tile, and the debug/dev/diag chips (build provenance, font-size cycle,
  hold-to-repeat, force-touch) — each gated by #135's tier classes.
- Touch panes: a 3×3 D-pad + persistent action column, three panes
  (MOVE / ITEMS / KEYS) cycled by the nav tile, hold-to-repeat on the
  directions, pointer→`sendInput` on the same path as xterm keystrokes.
- Game-event wiring (`xsofy/startup`, `/stats`, `/term-size`, `/perf`) driving
  the self-revealing diagnostics, plus the runtime OPTIONS and force-touch
  toggles.

Terminal core stays #154's (letterbox refit, font-await, BEL nudge). In dev the
build chip's `build-info.json` 404s in the console — deploy generates it, so
it's absent locally; harmless.

#135 + this = the old #135's full touch shell, forward-ported onto #154. The
landscape split follows on top.

### DIFF
--- a/tools/xsofy-shell.html
+++ b/tools/xsofy-shell.html
@@ -128,6 +128,112 @@ body.force-touch #xsofy-shell .touch { display: grid; }
 #xsofy-shell:not(.mode-urldiag) .diag-only  { display: none; }
 #xsofy-shell:not(.mode-dev)     .dev-only   { display: none; }
 #xsofy-shell.mode-debug .play-only          { display: none; }
+
+/* ============================================================
+   controls + style — opinionated chrome layered onto the scaffold above.
+   Theming, the info-bar grid, and the button/tile look. These rules only add
+   appearance to the structure already established; they don't restructure it.
+   ============================================================ */
+#xsofy-shell {
+  --bar-bg: #110d08;
+  --bar-fg: #ede1c4;
+  --bar-dim: #7a6d52;
+  --bar-accent: #f5b041;
+  --bar-rule: rgba(245, 176, 65, 0.18);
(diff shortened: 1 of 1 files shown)


mparrett:
Reviewer note: this slice is the foundation (theming, info bar, event wiring, press/repeat machinery) and reads best together with #137 stacked on it. A few things introduced here are superseded up-stack — the OPTIONS ⚙↔← cycle tile and the info-bar option chips become a 2-segment switcher + config pane in #137 — so don't invest in polishing those; the plumbing underneath them is what lands.

mparrett:
Folded into #137. After this morning's rebase, #137's branch already contained this PR's commit, so #137 simply retargeted onto shell/app-mount and now shows the complete mobile shell as one diff — where the OPTIONS tile this slice introduced (and the redesign retired) nets out entirely, making the earlier reviewer note moot. Review continues on #137.

PR #156, opened 2026-07-12
feat(mobile): touch relayout — 4-row grid, mirrored, persistent y/n + keyset row

mparrett:
**Draft / test-only.** Stacked on #137 so the deploy preview shows the full touch UI (portrait + landscape) with the relayout. Not for independent merge — **folds into #155** if approved.

Two changes to `tools/xsofy-shell.html`:

- Mirror the `.touch` grid — action column to the **left** (2fr), D-pad to the **right** (3fr), via explicit `grid-column` placement (no DOM reorder).
- Reorder `RIGHT_COLUMN` so **Esc/BACK leads** → top-left of the (now left) action block, matching Esc's spot on a QWERTY keyboard.

Preview will publish once the `deploy-preview` build finishes.

---
Screenshots

Laptop (Chrome; responsive size/aspect)

chrome-laptop-responsive

Laptop (Chrome; full size/aspect)
image

### DIFF
--- a/tools/xsofy-shell.html
+++ b/tools/xsofy-shell.html
@@ -23,6 +23,12 @@
      Fairfax HD has no Reserved Font Name, so the subset keeps the family name.
      License: tools/fonts/LICENSE-FairfaxHD-OFL.txt. -->
 <style>
+/* Own the color scheme so iOS Safari doesn't dark-mode-transform our fixed dark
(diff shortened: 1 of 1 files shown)


mparrett:
Folded into #137. The relayout grew into the full touch-shell redesign (portrait + landscape), and reviewing the split and relayout as separate intermediate states stopped making sense; mobile/landscape now carries the whole series. See #137 for the updated description.

PR #158, opened 2026-07-13
feat(input): bind Plan 9 arrow-key runes (fixes down-arrow in alacritty9)

### LINKED ISSUE #157: Plan 9: arrow keys arrive as kernel runes (down-arrow dead in alacritty9)
Whether arrow keys reach a program on Plan 9 — and in what form — depends entirely on the terminal. This came out of getting xsofy running on 9front (via let-go's Plan 9 input work, nooga/let-go#461). Movement already works everywhere through **hjkl + numpad**; this issue is specifically about the arrow keys, which behave differently in each environment.

## Arrow keys by terminal

| Terminal | Arrow keys | Renders ANSI? |
|---|---|---|
| **alacritty9** (agent9) | arrive as ANSI `ESC[A`/`[B`/`[C`/`[D` — xsofy's existing bindings already handle them | yes |
| **rio native window** (`rc`, `vtwin`) | rio intercepts them for scrollback (`Kup`/`Kdown` are its view/scroll keys); they never reach the program, even in raw mode. Letters (hjkl) do reach | no |
| **bare console** (no rio) | reach the program as Plan 9 key runes (see below), delivered by the kernel keyboard driver | no |

The upshot: **alacritty9 is where you actually play** (it renders xsofy's ANSI and its arrows already work). rio windows can't render the game's output at all, and additionally swallow the arrow keys. The bare console delivers arrow runes to the program but also can't render ANSI.

## Correction to the original framing

An earlier version of this issue said rio "sends" these runes. That's wrong: in a rio window rio *eats* the arrow keys for scrollback, so a program never sees them. The runes only reach a program on the **bare** console, and it's the kernel keyboard driver delivering them — not rio.

## The bare-console runes

From `/sys/include/keyboard.h`:

| Arrow | Name | Codepoint |
|---|---|---|
| Up | `Kup` | U+F00E |
| Down | `Kdown` (= `Kview`) | U+F800 |
| Left | `Kleft` | U+F011 |
| Right | `Kright` | U+F012 |

## Proposed change (low priority)

Binding these four runes in `base-key-bindings` movement `:keys` in `xsofy/input.lg` makes the arrow keys move the player **on the bare console**. Additive — ANSI terminals and vi-keys are untouched. A branch is ready.

But its value is narrow: the bare console is the *only* place it helps, and the bare console can't render xsofy's ANSI anyway, so it's a poor play environment. In alacritty9 — the real target — arrows already work via `ESC[A`, and hjkl/numpad work everywhere. So this is a completeness nicety for the bare console, not a blocker for Plan 9 playability.

Keeping the issue as documentation of the arrow-input landscape; happy to land the runes if we want the bare console fully covered.

mparrett:
## What

Movement keys bound only the ANSI arrow escapes (`ESC[A/B/C/D`). This adds Plan 9's kernel arrow runes (Kup `U+F00E`, Kdown `U+F800`, Kleft `U+F011`, Kright `U+F012`) alongside them, so the arrow keys drive movement on the Plan 9 terminals that deliver runes instead of escapes.

## Why it's not cosmetic

Which form a terminal sends isn't uniform, and the gap lands on the main play environment:

- **alacritty9** — where the game is actually played on 9front — sends ANSI for up/left/right, but the raw **Kdown rune `U+F800`** for down. Measured on 9front amd64: without the `U+F800` binding, up/left/right move and **down does nothing**.
- The **bare text console** sends all four keys as runes.
- rio's own windows swallow the arrows for scrollback, so they never reach the program — not something a key binding can address.

The down case is the one that bites: a silent dead key in the primary environment, easy to miss because hjkl and the numpad still move you.

## Evidence

Probed on real hardware (9front amd64 + alacritty9) with a `read-key` codepoint dump, entering the alternate screen the way the game does:

```
↑  codes=(27 91 65)  ESC[A
↓  codes=(63488)     U+F800   ← Kdown rune, every press
←  codes=(27 91 68)  ESC[D
→  codes=(27 91 67)  ESC[C
```

The let-go side (`read-key` tokenizing these on Plan 9) is nooga/let-go#461; this is the xsofy-side binding that consumes them.

## Test

`input_test.lg` covers the rune bindings resolving to the right movement actions.

Closes #157.

### DIFF
--- a/xsofy/input.lg
+++ b/xsofy/input.lg
@@ -11,11 +11,19 @@
   (or (= (js/url-param "mode") "dev")
       (let [e (os/getenv "XSOFY_DEV")] (and e (> (count e) 0)))))
 
+;; Movement :keys carry both the ANSI arrow escapes and Plan 9's kernel arrow
+;; runes (Kup U+F00E, Kdown U+F800, Kleft U+F011, Kright U+F012 —
+;; /sys/include/keyboard.h). Which form a terminal delivers varies, so we bind
+;; both: alacritty9 sends ANSI ESC[ for up/left/right but the raw Kdown rune for
+;; DOWN (measured on 9front — down-movement in-game needs U+F800 or it's dead);
+;; the bare text console sends all four as runes; rio's own windows swallow the
+;; arrows for scrollback. Binding both forms makes movement work wherever a key
+;; actually reaches us.
 (def base-key-bindings
-  [{:action :up          :category :movement   :label "Move up"          :keys ["\u001b[A" "k" "8"] :help-keys ["k" "8" "Up"]}
-   {:action :down        :category :movement   :label "Move down"        :keys ["\u001b[B" "j" "2"] :help-keys ["j" "2" "Down"]}
-   {:action :left        :category :movement   :label "Move left"        :keys ["\u001b[D" "h" "4"] :help-keys ["h" "4" "Left"]}
-   {:action :right       :category :movement   :label "Move right"       :keys ["\u001b[C" "l" "6"] :help-keys ["l" "6" "Right"]}
+  [{:action :up          :category :movement   :label "Move up"          :keys ["\u001b[A" "\uf00e" "k" "8"] :help-keys ["k" "8" "Up"]}
+   {:action :down        :category :movement   :label "Move down"        :keys ["\u001b[B" "\uf800" "j" "2"] :help-keys ["j" "2" "Down"]}
+   {:action :left        :category :movement   :label "Move left"        :keys ["\u001b[D" "\uf011" "h" "4"] :help-keys ["h" "4" "Left"]}
+   {:action :right       :category :movement   :label "Move right"       :keys ["\u001b[C" "\uf012" "l" "6"] :help-keys ["l" "6" "Right"]}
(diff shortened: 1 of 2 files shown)


mparrett:
## Bare-console confirmation: all four runes on real hardware

The PR body measured only the alacritty9 down-arrow; the other three rune values came from `keyboard.h`. Ran the arrow probe on a bare 9front console (rio exited, raw `/dev/cons`) and all four match the bindings:

```
↑  codes=(61454)  U+F00E  Kup
↓  codes=(63488)  U+F800  Kdown
←  codes=(61457)  U+F011  Kleft
→  codes=(61458)  U+F012  Kright
```

(`raw=""` because they're Private-Use-Area codepoints with no glyph — the codepoints are the receipt.) Each arrived unbuffered, one rune per press, under consctl raw mode; `q` quit cleanly with the console back in cooked mode.

So the two device paths together cover every key this binds: the bare console sends all four as runes, and alacritty9 sends ANSI for up/left/right plus the Kdown rune for down.

mparrett:
Just a hunch, this PR seems like it's trying to fix something that shouldn't be that way in the first place. The "raw Kdown rune U+F800 for down" part smells like a bug rather than something to accommodate here- Not proven, but need to take a deeper look.

PR #71, opened 2026-06-07
balance: tame slimes, throttle ranged fire rate, soften spider webs

nooga:
## What

Tuning pass on early-dungeon difficulty. Three creature families were disproportionately lethal:

- **Slimes** were the standout killer — tanky (50 HP), fast (`speed 2` = acts *twice* per player turn), **and** splitting on every hit. The triple-combo snowballed out of control.
- **Every ranged enemy except the spider** omitted `:fire-chance`, which `hunt-ranged` defaulted to **100%** — so archer/goblin-shaman/fire-imp/ice-elemental/dragon all shot every single turn.
- **Spiders** acted twice per turn (`speed 2`) and rooted the player with `webbed 3`, keeping you in the kill zone while they kept firing.

## Changes

| Creature | Change | File |
|---|---|---|
| Slime | HP `50→24`, speed `2→1` (keeps splits-on-hit) | `xsofy/bestiary.lg` |
| All ranged | New `default-ranged-fire-chance` constant = `60` (was hardcoded 100% default) | `xsofy/ai.lg` |
| Spider | speed `2→1` | `xsofy/bestiary.lg` |
| Player web-root | `webbed 3→2` on stepping into a web (enemy-side `:web` rune potency left at 3) | `xsofy/world.lg` |

## Measured effect

60-run seeded balance sweep (`starter` loadout, `greedy-melee` policy, seed 1000):

| Metric | Before | After |
|---|---|---|
| Deaths | 51/60 | 42/60 |
| Mean turns survived | 158 | 288 |
| slime kills | 18 | 4 |
| spider kills | 11 | 4 |
| archer kills | 9 | 14 |

Slime and spider kills both dropped ~70%; overall survival nearly doubled. Archer kills rose, but that's a survivorship + test-bot artifact: players now survive the early game and reach more archers, and the melee-only sim bot has no ranged counterplay (no kiting/cover). Per-shot archer lethality genuinely dropped (60% vs 100%).

## Testing

- Full test suite passes (2233 assertions).
- `balance.lg` is a pure runtime sim harness — no hardcoded stats, nothing to break.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #166, opened 2026-07-21
fix(status): stop generic decay from racing the web struggle branch

### LINKED ISSUE #161: Spider webs no longer trampled
The spider webs don't disappear when a creature breaks free. This seems like a regression.

mparrett:
Closes #161. Full walkthrough of the mechanism is in the issue comment; short version: `:webbed` ttl was decremented twice per turn — once by the movement struggle branches (the only place that tramples, at ttl ≤ 1) and once by the generic status tick (which removes an expired status silently). The status always expired inside the tick, so the trample branch was unreachable. #71's ttl reduction (3 to 2) exposed it on the player path; the monster path (tick before action in `creature-turn`) had never trampled since webs landed.

Note: Temporarily changed ash from . to , in order to disambiguate and verify the changes. Not included in this PR, but might be worth discussing when we get back to gameplay aesethetics.

image

## Approach

Statuses can declare `:no-decay` in `status-defs`, and `:webbed` does: the generic tick skips it, and the struggle branches become the sole owner of the ttl. One mechanism fixes both paths, and the ttl now means what the #71 balance knob assumed: the number of struggles to break free. The alternative — trampling on expiry inside the tick — would have coupled `status.lg` to `fire.lg` and kept two owners of the same counter.

Behavior note (balance-visible): a webbed creature that never attempts a move — say, attacking an adjacent target — now stays webbed until it does, instead of timing out of the web by standing still. That reads as correct to me; easy to revisit if it plays badly.

## Tests

New `xsofy/test/status_test.lg`: the generic tick leaves `:webbed` untouched while `:confused` still decays, and a player struggling free ends on ash, not an intact web. Suite is green (299 tests, 2777 assertions); the new assertions fail on main.

### DIFF
--- a/xsofy/test/status_test.lg
+++ b/xsofy/test/status_test.lg
@@ -1,0 +1,40 @@
+(ns xsofy.test.status-test
+  (:require [test :refer [deftest is testing]]
+            [xsofy.terrain :as terrain]
+            [xsofy.fire :as fire]
+            [xsofy.status :as status]
+            [xsofy.world :as world]))
+
+;; Regression tests for #161: webs no longer trampled.
+;;
+;; :webbed ttl is decremented by the movement struggle branches, which
+;; trample the web on the last struggle. The generic status tick also
+;; decremented it, so the ttl drained twice per turn and expired inside
+;; the tick — which removes the status silently, without trampling. The
+;; struggle branch then never saw ttl <= 1 and the web survived every
+;; escape. :webbed is now :no-decay: the struggle branches own its ttl.
+
+(deftest generic-tick-leaves-webbed-alone
+  (testing ":no-decay statuses keep their ttl through the generic tick"
+    (let [w {:entities {:e {:id :e :pos [1 1]}}}
+          w (status/apply-status w :e :webbed 2)
+          w (status/apply-status w :e :confused 2)
+          w (status/tick-entity-statuses w :e)]
+      ;; still webbed, ttl untouched
+      (is (= 2 (get-in w [:entities :e :statuses :webbed :ttl])))
+      ;; the normal decay path still works
+      (is (= 1 (get-in w [:entities :e :statuses :confused :ttl])))
+      (let [w (status/tick-entity-statuses w :e)]
+        (is (status/has-status? w :e :webbed))
+        (is (not (status/has-status? w :e :confused)))))))
+
+(deftest webbed-player-break-free-tramples-web
+  (testing "player struggling free of a web leaves ash, not an intact web"
+    (let [w (world/make-world 30 20 1)
+          [px py] (get-in w [:entities :player :pos])
+          _ (terrain/tset! (:terrain w) (:width w) px py fire/web-tile)
+          w (status/apply-status w :player :webbed 2)
+          w (world/player-move w 1 0)   ;; struggle: ttl 2 -> 1
+          w (world/player-move w 1 0)]  ;; final struggle: break free + trample
+      (is (not (status/has-status? w :player :webbed)))
+      (is (= fire/ash-tile (terrain/tget (:terrain w) (:width w) px py))))))
--- a/xsofy/status.lg
+++ b/xsofy/status.lg
@@ -6,6 +6,8 @@
 ;;   :ttl       — turns remaining
 ;;   :dmg       — damage per tick (optional)
 ;;   :skip-turn — entity can't act (optional)
+;;   :no-decay  — ttl is owned by a dedicated system; the generic tick
+;;                leaves it alone (optional)
 ;;   :color     — tint for rendering
 
 (def status-defs
@@ -15,7 +17,11 @@
    :confused {:color [200 100 255] :bg [40 20 50] :name "confused"}
    :slowed   {:color [140 140 180] :bg [30 30 40] :name "slowed"}
    :haste    {:color [255 255 100] :bg [40 40 15] :name "haste"}
-   :webbed   {:color [200 200 210] :bg [30 30 32] :name "webbed"}
+   ;; :webbed ttl counts remaining struggles; the movement struggle branches
+   ;; (world/player-move, ai/try-move) decrement it and trample the web on
+   ;; the last one. Generic decay raced them and freed the creature with the
+   ;; web intact (#161), so the tick must not touch it.
+   :webbed   {:color [200 200 210] :bg [30 30 32] :name "webbed" :no-decay true}
    :swimming {:color [60 100 180] :bg [10 20 50] :name "swimming"}})
 
 ;; --- Apply / Remove ---
@@ -51,8 +57,10 @@
       world
       (reduce
         (fn [w [stype sdata]]
-          ;; Bail out if a prior tick (or external code) removed the entity.
-          (if (nil? (get-in w [:entities entity-id]))
+          ;; Bail out if a prior tick (or external code) removed the entity,
+          ;; or if the status decays through its own system (:no-decay).
+          (if (or (nil? (get-in w [:entities entity-id]))
+                  (:no-decay (get status-defs stype)))
             w
             (let [def_ (get status-defs stype)
                   ttl (dec (or (:ttl sdata) 1))
--- a/xsofy/test/run.lg
+++ b/xsofy/test/run.lg
@@ -15,6 +15,7 @@
             [xsofy.test.dag-test]
             [xsofy.test.container-trigger-test]
             [xsofy.test.fire-test]
+            [xsofy.test.status-test]
             [xsofy.test.seed-test]
             [xsofy.test.mapgen-audit-test]
             [xsofy.test.replay-test]


mparrett:
FYI — initia self-review notes.

**Root cause confirmed.** `:webbed` ttl was decremented by two independent systems — the movement struggle branches (`world/player-move`, `ai/try-move`) and the generic per-turn status tick — so it drained twice a turn and expired *inside* the tick, which removes the status silently and never tramples the web. Marking `:webbed :no-decay` hands sole ownership to the struggle branches, where the trample lives. Right fix, narrow scope.

**Validated.** Full suite green (299 tests / 2777 assertions), including `replay-test`, so determinism holds. Differential check: with `main`'s pre-fix `status.lg` swapped back in, both new tests fail (webbed drains early, web left un-trampled) and pass with the fix — real guards.

**One behavior change to decide on:** breaking free now requires a move *attempt*. Resting/waiting no longer decays the web, and a webbed monster that only sleeps or attacks in place stays webbed indefinitely. This matches the documented intent (`fire.lg:136` — "after N struggles the web is trampled"), so waiting-out was probably the accident. Open question: should the web still tick down when you wait?

**Minor:** `:no-decay` short-circuits the whole status entry, including `:dmg`. Harmless for `:webbed` (no damage), but a future damaging `:no-decay` status wouldn't tick — worth a comment at the guard.

Not a recent regression: the double-decrement dates to the original web commit (`8feaa1c`), not a later break.

mparrett:
Tested web struggle and it behaved predictablly.

image

mparrett:
> Tested web struggle and it behaved predictablly.
> 
> image

not seen (well) in this image: ash.
Added a new screenshot to the PR body that shows it, unambiguously.

mparrett:

xsofy/test/status_test.lg
```diff
@@ -0,0 +1,40 @@
...
+
+;; Regression tests for #161: webs no longer trampled.
+;;
+;; :webbed ttl is decremented by the movement struggle branches, which
+;; trample the web on the last struggle. The generic status tick also
+;; decremented it, so the ttl drained twice per turn and expired inside
+;; the tick — which removes the status silently, without trampling. The
+;; struggle branch then never saw ttl <= 1 and the web survived every
+;; escape. :webbed is now :no-decay: the struggle branches own its ttl.
+
```
nnunley, line 16: This feels like the wrong solution. I'd suggest digging into the root cause - updates for things like decay should only happen once at the end of the loop - shouldn't everything more or less be considered to be simultaneous during a tick?

This would suggest there are other things that are either order dependent or are mutating state multiple times instead of the single time we're expecting.

xsofy/test/status_test.lg
```diff
@@ -0,0 +1,40 @@
...
+(deftest generic-tick-leaves-webbed-alone
+  (testing ":no-decay statuses keep their ttl through the generic tick"
+    (let [w {:entities {:e {:id :e :pos [1 1]}}}
+          w (status/apply-status w :e :webbed 2)
+          w (status/apply-status w :e :confused 2)
```
nnunley, line 21: Why is :confused being touched here?

nnunley (dismissed):
Sorry for the delay. I think this is probably the wrong solution for this problem. I think that there should be a consolidation pass that's idempotent for the effects of actions within the turn (order of application should preferably not matter within the tick). The web decay's tick should be calculated as a forward projected time in rounds/ticks

Note that a player turn should take  some multiple > 1 worth of ticks (like maybe 1000) to allow spells like haste to actually do something interesting. Trampling should have at most one entity affecting the object, and should have a handler for not just webs, but grasses, water, etc.

I'm willing to let this pass, but there feels like some hackery going on to get things working.

mparrett:
> Sorry for the delay. I think this is probably the wrong solution for this problem. I think that there should be a consolidation pass that's idempotent for the effects of actions within the turn (order of application should preferably not matter within the tick). The web decay's tick should be calculated as a forward projected time in rounds/ticks
> 
> Note that a player turn should take  some multiple > 1 worth of ticks (like maybe 1000) to allow spells like haste to actually do something interesting. Trampling should have at most one entity affecting the object, and should have a handler for not just webs, but grasses, water, etc.
> 
> I'm willing to let this pass, but there feels like some hackery going on to get things working.

Thanks for the great feedback. Let me take a fresh look and see if it makes sense to move this forward or revise.

xsofy/test/status_test.lg
mparrett: `:confused` is the control. The pair asserts that the generic tick leaves `:webbed` alone *while `:confused` still decays*, so the test fails if `:no-decay` ever switches decay off globally instead of for the flagged status. I can rename it to make that intent obvious.

xsofy/test/status_test.lg
mparrett: I'll respond fully in the main thread in a moment.

Short version: I checked for the other double-mutations and `:webbed` is the only status with two owners, but the tick model is inconsistent as you suspected. Status ttl decrements per
_action_ rather than per _tick_, so a swim that costs 2x burns the same ttl as a step. More detail and suggested sequencing coming shortly.

mparrett:
@nnunley Thanks again for the review. Here's the longer response after looking through the code with Claude.

I tested the "mutating state multiple times" half of your point and it doesn't hold for the per-turn counters: just two places outside `status.lg` write a status ttl, and both are the `:webbed` struggle branches relevant to this PR:

1. the player branch in `player-move`
2. the creature branch in `ai.lg`

Every other status goes through `apply-status` / `remove-status` (11 sites) and the single `tick-entity-statuses`. The surrounding turn machinery is clean: `update-world` dispatches one action per call, every `add-ticks :player` site sits in a mutually exclusive branch, `tick-vfx` has one caller, and `step-fire`, `tick-pending-allies`, `burning-entities-ignite`, and `drift-water-items` appear twice in `world.lg` but only because they are duplicated across the two arms of `(if *perf-detail* ...)`, a bench-timing split with the same sub-phase order. So `:webbed` is the only counter with two owners.

I didn't test the order-dependence half — that's a different property, and single-application doesn't imply order-independence.

**On the tick model**

I think you're right: `add-ticks` scales `:ticks` and `regen-ticks` by `cost`, then unconditionally calls `tick-entity-statuses`. As a result, status ttl decrements once per action and is never scaled by what that action cost. Swimming costs `(* move-cost 2)` and burns one ttl, the same as a normal step; a speed-modified creature decays at a rate decoupled from the ticks it actually spends. ttl is documented as "turns remaining" and is really "actions remaining". Your intuition about a forward-projected expiry makes sense in this context, and doesn't depend on the web bug.

It's also what stands between us and your 1000-ticks-per-turn idea: haste can't mean much while decay counts actions rather than ticks.

And we have this asymmetry: player ticks statuses inside `add-ticks`, creatures tick at the top of `creature-turn` before acting. One concept, two homes, which is how a second owner got in.

Projected time would be the wrong fit for `:webbed`. Its ttl is not a status effect timer, but a counter of struggles to break free, per #71. The behavior note in the description says: a webbed creature that never attempts a move stays webbed. Convert it to projected ticks and a web times out while you stand still. That's a different rule, and we should consider it, but it's deliberately out of scope here. A consolidation pass likely makes sense, but it has to model counters as well as timers, or `:webbed` becomes a timer on the way through.

So unless the above changes your mind, I'll merge this as the narrow fix, since #161 is a live regression and the new assertions fail on `main`. I'll then open an issue for the consolidation pass carrying the per-action-versus-per-tick finding above, which has a larger blast radius and needs some design thinking — and I'll fold your trample point into it, since "at most one entity affecting the object" and a handler covering grasses and water as well as webs is the same shape of problem.

That said, if you prefer `:no-decay` isn't in the tree at all, I'm happy to close this in favor of the refactor, which will then take responsibility for addressing #161.

mparrett:
One amendment to my own survey. The order-dependence half is already tracked: #187 is that case. `process-creatures` (`ai.lg:424`) takes creature acting order straight from `(vals (:entities world))` with no sort, and since every creature action advances the `det` seed chain, acting order is part of the determinism contract. `find-target` already guards the same dependency by sorting on `(juxt dist-to (comp str :id))`; `process-creatures` doesn't.

I filed that on 2026-08-20, before your review, so your instinct had more behind it than my "I didn't test that half" line suggested. It goes into the consolidation-pass issue as the order half.

nooga (approved):
Reviewed via automated multi-angle analysis (line-by-line, removed-behavior, cross-file trace, cleanup, efficiency, altitude, CLAUDE.md conventions) plus a full test-suite run.

**Verdict: approve.** The fix is correct — `:no-decay` cleanly exempts `:webbed` from the generic per-turn tick so it's only cleared by the struggle-resolution branches, closing the double-decrement race. Full suite passes (318 tests / 2825 assertions), including the two new regression tests.

Two non-blocking notes for a follow-up, not this PR:

1. `xsofy/status.lg:63` — `tick-entity-statuses` looks up `status-defs` twice per status per tick (once for the `:no-decay` check, again a couple lines later as `def_`). Minor hot-path inefficiency; hoisting the lookup above the `if` would remove the redundant map lookup.
2. `xsofy/status.lg:24` — `:webbed` now has no time-based expiry at all; it only clears via the struggle branches, which only fire on a move attempt. An entity that never moves again (docile AI, or a player who only attacks/waits) stays webbed indefinitely. This is likely intended ("only struggling frees you"), but it's a real behavior change from before and isn't called out in the docstring/comment — worth a one-line note.

PR #167, opened 2026-07-22
shell: add 'off' to the force-touch cycle (auto/force/off)

mparrett:
**This PR is about the shell's touch UI toggle.** (here referenced as "force-touch" since it originally forced the UI on in layouts where it would normally be invisible).

image

The ▦ dev toggle in the shell initially was a boolean: `force` (touch UI pinned on) or `auto` (pointer-capability detection). I discovered in testing in desktop browers, it is helpful to have an explicit `off` option. The current `auto` isn't necessarily `off` — portrait's media query and a coarse pointer both reveal the touch UI regardless, so there was no way to suppress it when a reveal path fires but the UI isn't wanted: a phone with a Bluetooth keyboard, or a **narrow/portrait desktop window**.

This PR adds a natural third state to the cycle: `auto` → `force` → `off`.

**How:** `off` sets `body.no-touch`, which beats both reveal paths for `.touch` on plain specificity (no `!important`) and releases the portrait `55dvh` terminal cap, so the grid refits to the full viewport.

**Recoverability:** the switcher and settings pane stay visible under `mode-dev` only, the same contract that already kept ▦ reachable on a plain desktop. A dev can always cycle back out; players never see the toggle, so they can't strand themselves in a touch-less state on a phone.

The persisted value moves from `'1'`/`'0'` to the mode name; a stored legacy `'1'` maps to `force` on load, so nobody's existing override resets. (mostly a non-issue for anyone except me right now, but minions love to defensively add backward compat for everything. c'est la vie)

Validated manually and with Playwright in a portrait viewport, tapping through the full cycle and back to `auto`: in `off` the touch grid hides, the terminal grows from 495px to 847px and refits, and the `mode-dev` switcher remains as the way back.

🤖 Co-authored with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tools/xsofy-shell.html
+++ b/tools/xsofy-shell.html
@@ -240,6 +240,21 @@ body.force-touch #xsofy-shell.show-settings .settings { display: flex; }
 @media (hover: hover) and (pointer: fine) {
   body:not(.force-touch) #xsofy-shell .touch-only { display: none; }
 }
+/* Touch mode 'off' (third state of the ▦ cycle): suppress the touch UI even
+   where portrait or a coarse pointer would reveal it — phone with a BT
+   keyboard, narrow desktop windows. The terminal reclaims the viewport
+   (overrides both the portrait 55dvh cap and portrait's flex:1 shell). The
+   mode-dev re-enables keep the switcher + settings pane reachable: that's the
+   only way back out of 'off', same recoverability contract as line "dev tier
+   gets the switcher" above. no-touch and force-touch are mutually exclusive
+   (the JS sets at most one). */
+body.no-touch #xsofy-shell .touch { display: none; }
+body.no-touch #xsofy-shell .seg-switch { display: none; }
+body.no-touch #xsofy-shell.mode-dev .seg-switch { display: inline-flex; }
+body.no-touch #xsofy-shell .settings { display: none; }
+body.no-touch #xsofy-shell.mode-dev.show-settings .settings { display: flex; }
+body.no-touch #app { flex: 1 1 auto; height: auto; }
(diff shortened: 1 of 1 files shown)


mparrett:

nnunley:
Perhaps the better UI would be a dropdown selection for the mode, rather than toggle cycling. Having tooltips for what each do would be helpful, too. 

On chrome, I'm unable to enter the console still. It's ` right?

mparrett:
> Perhaps the better UI would be a dropdown selection for the mode, rather than toggle cycling. Having tooltips for what each do would be helpful, too.
> 
> On chrome, I'm unable to enter the console still. It's ` right?

@nnunley Thanks for taking a look and agree there's better UI. I'll capture this dimension with or along with the other UI spikes.

And yeah, that's the right key. Add `?mode=dev`. Doesn't activate until after first descent. https://xsofy.quest/pr-preview/pr-167/?mode=dev

mparrett:
FYI, there are a few notes left over here: https://github.com/nooga/xsofy/issues/182 in case it helps with testing.

mparrett:
> Perhaps the better UI would be a dropdown selection for the mode, rather than toggle cycling. Having tooltips for what each do would be helpful, too.
> 
> On chrome, I'm unable to enter the console still. It's ` right?

@nnunley Curious if you ever got around to trying the console experience with the `?mode=dev` link above? And I agree about eventual drop-down as a preferred UX (especially on desktop). The current toggle is admittedly biased toward mobile, but works across both for now and I'd argue has a smaller implementation surface. That said, any more extensive menu-driven interaction and we'll probably want something like what you're describing. Ok with deferring until then?

nnunley (approved):

PR #168, opened 2026-07-22
demo(shell): WebGL renderer + post-process shader overlay (CRT, water, mode7) - DEMO/POC

mparrett:
This is a proof of concept spawned  by wondering, since xterm has an opengl canvas, can we use it, and then do fun shader stuff in opengl?

I turns out we _can_. I am not sure we _should_. Some concepts implemented here we've discussed before: game state emits to js, triggers (for e.g. sound effects), and might be worth extracting separate from the shader experiments.

But this is fun to look at. Let me know what you think!

**A few demo shaders:**

**water**

https://github.com/user-attachments/assets/08063db0-faca-4f3d-a631-503a826a73de

**mode7**

https://github.com/user-attachments/assets/3012baf2-5242-4979-b4b2-ac0548baf02a

**crt**

https://github.com/user-attachments/assets/25037219-0cd5-4cc1-88a7-e99b48e82150

---

Draft/spike for discussion — the terminal gets a WebGL renderer and a shader post-process pass, in two layers that stand alone:

**(a) WebGL renderer.** `@xterm/addon-webgl` from the same jsdelivr lane as xterm/fit, loaded after `term.open()`, with a context-loss fallback to the DOM renderer. `?renderer=dom` opts out.

**(b) Shader overlay.** A WebGL2 canvas re-samples the renderer's canvas every frame through a Shadertoy-convention wrapper (`iChannel0`/`iResolution`/`iTime` + `mainImage`), so any Shadertoy-style shader drops in. Aligned to the terminal canvas, `pointer-events: none`, tracking letterbox refits per frame. `?crt=off` opts out while keeping (a).

Four shaders in the registry, selected by `?shader=`; programs compile lazily and can switch at runtime. Default is none — the page starts with no shader applied (and the overlay sits dormant, costing nothing) until `?shader=`, a game signal, or a console override activates one:

- `crt` — Timothy Lottes' CRT filter (public domain / Unlicense, as adapted for Ghostty): scanlines, shadow mask, tube warp, vignette. The filter re-scans the input at `iResolution × SCALE`, so SCALE below 1/dpr costs text sharpness; the default anchors at CSS resolution (0.5 on retina, 2/3 at dpr 1), `?crtscale=0.2..1` to taste.
- `water` — animated caustic ripple.
- `bloom` — 24-tap golden-spiral glow, from the same qwerasd gist as the CRT adaptation.
- `mode7` — SNES-style affine spin/zoom demo, sampled nearest-neighbor on a tiled plane (registry entries choose their sampling: NEAREST and REPEAT here, LINEAR and CLAMP for the others). Idle-animated; wiring it to game transitions (descend swirl) would ride the signal channel below plus a map-rect uniform so the HUD stays put (not built).

**Game-signaled effects.** The game emits a private OSC escape (`7777`, `k=v` pairs) before frame flush, only when the signal set changes. Real terminals discard unknown OSC (native TUI unaffected), and the emission is output-only, so determinism/replay are untouched. Wired so far: the `:swimming` status swaps the active shader to `water` for the duration, and health below 35% drives a Doom-style red pulse the wrapper composes over whichever shader is active. The dev console can drive the same channel: `(crt)` / `(water)` / `(bloom)` / `(mode7)` toggle a shader override (repeat toggles off), `(hurt N)` forces the pulse, `(pulse rune)` / `(pulse item)` fire one-shot pickup flashes (purple / white — transient additive washes with a ~500ms decay; a plain passthrough wakes the dormant overlay so they show even with no shader selected; in-game pickup emits are a follow-up), `(fx)` clears. The console stays read-only: these are output-only escapes. Precedence: console override, then swim, then the URL base.

To try it on the deploy preview: open [xsofy.quest/pr-preview/pr-168/?mode=dev](https://xsofy.quest/pr-preview/pr-168/?mode=dev) (the `mode=dev` param gates the console), then press `` ` `` (backtick) in-game to open it and type `(help)` for the command list.

**Contrast.** Some effects darken the scene (water's caustic multiply bottoms out around 0.6x), so the wrapper compiles a per-shader mid-anchored contrast stretch + brightness lift after `mainImage` (water ships 1.2/+0.1); `?contrast=` / `?bright=` override any shader for tasting.

**Implementation notes:**

- The addon needs `preserveDrawingBuffer: true` — the overlay samples outside the addon's rAF, and a non-preserved backbuffer reads back blank after compositing.
- `.xterm-screen` holds two canvases; the first is the transparent link-underline layer, which samples as solid black. The overlay selects `:not(.xterm-link-layer)`.
- The per-frame `texImage2D` from the renderer canvas is a GPU readback (Chrome logs ReadPixels stall warnings). Fine at terminal sizes in testing; a shared-context or OffscreenCanvas route would remove it if this graduates.
- The WebGL renderer never populates `.xterm-rows` with DOM text, which silently blinds DOM-scraping assertions; the PR-preview smoke gate failed exactly this way. The shell now exposes `window.xsofyTermText` (a renderer-independent readout via xterm's buffer API) and the smoke prefers it, falling back to DOM scraping for builds without it.

**For discussion**

- `water.glsl` came from the ghostty-shaders collection without a license header (it's the widely-copied Shadertoy caustic loop) — provenance needs chasing or the shader swapped/dropped. `crt` and `bloom` are clean (same Unlicense gist).
- Whether shader selection should stay URL-only or get a settings-pane chip.
- Whether the WebGL renderer should default on, or ship behind the flag until it has more soak time (it changes the renderer for every player).

Validated with Playwright on the WASM bundle: renderer active, overlay tracking, all four shaders compiling and rendering, input unaffected; the game→shell signal path exercised end-to-end (boot emission reaches the handler; console commands drive the overrides); the CI smoke gate run locally against the WebGL build passes. Deploy preview is live for on-device poking.

### DIFF
--- a/tests/browser/smoke.mjs
+++ b/tests/browser/smoke.mjs
@@ -73,6 +73,10 @@ try {
 
   await page.waitForFunction(
     (sentinel) => {
+      // The WebGL renderer draws to a canvas atlas and never puts text in
+      // .xterm-rows, so prefer the shell's buffer-API readout when present;
+      // fall back to DOM scraping for builds without it (DOM renderer).
+      if (window.xsofyTermText) return window.xsofyTermText().includes(sentinel);
       const el = document.querySelector('.xterm-rows');
       return el && el.textContent.includes(sentinel);
     },
--- a/tools/xsofy-shell.html
+++ b/tools/xsofy-shell.html
@@ -680,8 +680,54 @@ void mainImage(out vec4 fragColor, in vec2 fragCoord) {
 }
 `;
 
+  // Animated water caustics — ripple-distorts the sampled terminal and layers
+  // caustic highlights; exercises the wrapper's iTime. From ghostty-shaders
+  // water.glsl (no license header there — chase provenance before upstreaming;
+  // it's the widely-copied Shadertoy caustic loop).
+  const WATER_GLSL = `
+#define TAU 6.28318530718
+#define MAX_ITER 6
+
+void mainImage( out vec4 fragColor, in vec2 fragCoord )
+{
+    vec3 water_color = vec3(1.0, 1.0, 1.0) * 0.5;
+    float time = iTime * 0.5+23.0;
+    vec2 uv = fragCoord.xy / iResolution.xy;
+
+    vec2 p = mod(uv*TAU, TAU)-250.0;
+    vec2 i = vec2(p);
+    float c = 1.0;
+    float inten = 0.005;
+
+    for (int n = 0; n < MAX_ITER; n++)
+    {
+        float t = time * (1.0 - (3.5 / float(n+1)));
+        i = p + vec2(cos(t - i.x) + sin(t + i.y), sin(t - i.y) + cos(t + i.x));
+        c += 1.0/length(vec2(p.x / (sin(i.x+t)/inten),p.y / (cos(i.y+t)/inten)));
+    }
+    c /= float(MAX_ITER);
+    c = 1.17-pow(c, 1.4);
+    vec3 color = vec3(pow(abs(c), 15.0));
+    color = clamp((color + water_color)*1.2, 0.0, 1.0);
+
+    // perterb uv based on value of c from caustic calc above
+    vec2 tc = vec2(cos(c)-0.75,sin(c)-0.75)*0.04;
+    uv = clamp(uv + tc,0.0,1.0);
+
+    fragColor = texture(iChannel0, uv);
+    // give transparent pixels a color
+    if ( fragColor.a == 0.0 ) fragColor=vec4(1.0,1.0,1.0,1.0);
+    fragColor *= vec4(color, 1.0);
+}
+`;
+
+  // Registry: ?shader=<name> picks one (default crt). Only the selected source
+  // is compiled, so the shaders can't collide with each other.
+  const FX_SHADERS = { crt: CRT_GLSL, water: WATER_GLSL };
+
   function startCrtOverlay() {
     if (!glAddon || fxParams.get('crt') === 'off') return;
+    const fxSource = FX_SHADERS[fxParams.get('shader')] || CRT_GLSL;
     const overlay = document.createElement('canvas');
     overlay.style.cssText =
       'position:fixed;left:0;top:0;pointer-events:none;z-index:10;display:none;';
@@ -706,7 +752,7 @@ void mainImage(out vec4 fragColor, in vec2 fragCoord) {
     const crtScaleLit = crtScale.toFixed(8); // GLSL float literal (never a bare int)
     const wrapFS = '#version 300 es\nprecision highp float;\n' +
       'uniform sampler2D iChannel0;\nuniform vec3 iResolution;\nuniform float iTime;\n' +
-      'out vec4 xsofyFragColor;\n' + CRT_GLSL.replace(/__CRT_SCALE__/g, crtScaleLit) +
+      'out vec4 xsofyFragColor;\n' + fxSource.replace(/__CRT_SCALE__/g, crtScaleLit) +
       '\nvoid main() { mainImage(xsofyFragColor, gl_FragCoord.xy); }\n';
     function compile(type, src) {
       const s = gl.createShader(type);
--- a/tools/xsofy-shell.html
+++ b/tools/xsofy-shell.html
@@ -800,7 +800,16 @@ void mainImage(out vec4 fragColor, in vec2 fragCoord) {
   // Shaders compile lazily on first activation, so entries can't collide
   // with each other. nearest/repeat control how the terminal texture is
   // sampled.
+  // Identity passthrough — what the frame loop activates when a transient
+  // effect (hurt, pickup pulse) needs the overlay but no shader is selected.
+  const PLAIN_GLSL = `
+void mainImage(out vec4 fragColor, in vec2 fragCoord) {
+  fragColor = texture(iChannel0, fragCoord / iResolution.xy);
+}
+`;
+
   const FX_SHADERS = {
+    plain: { src: PLAIN_GLSL },
     crt:   { src: CRT_GLSL },
     // water's caustic multiply darkens the whole scene; lift it back up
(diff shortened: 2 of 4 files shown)


mparrett:
Pushed a follow-up commit that answers the "how do effects trigger" question from the description: a game→shell signal channel over a private OSC escape (7777), demo-wired to two signals — the `:swimming` status swaps the active shader to the water effect, and health below 35% drives a Doom-style red pulse composed over whichever shader is active. Native terminals discard the unknown OSC, and the emission is output-only, so the TUI and determinism/replay are unaffected. Also adds `?shader=bloom` (same qwerasd gist as the CRT).

nooga (commented):
Reviewed via automated multi-angle analysis (JS/WebGL correctness and lifecycle trace, direct reproduction of one bug in Node, and a merge-conflict check against current `main`). This is explicitly a draft/spike, so treating this as feedback for the "let me know what you think" ask rather than a blocking review — nothing here should stop you from iterating, but flagging what I'd want fixed before it leaves spike status.

## Real, reproducible bug: prototype-chain lookups crash the overlay permanently

`FX_SHADERS`, `PULSE_KINDS`, and `progCache` in `tools/xsofy-shell.html` are plain `{}` object literals, and every lookup is a bare `obj[name]` truthy check — never `Object.hasOwn`. Any name colliding with an `Object.prototype` member (`constructor`, `toString`, `valueOf`, `hasOwnProperty`, `__proto__`) resolves to that inherited built-in and is treated as "found."

Concrete, no-injection-needed repro: load the preview with **`?shader=constructor`**. `FX_SHADERS['constructor']` is truthy (it's `Object`), so `ensureProgram('constructor')` finds `'constructor' in progCache` already `true` before anything compiles, returns the global `Object` function as the cache "entry," skips the try/catch, and `activate()` then does `entry.def.repeat` on `undefined` -> `TypeError`, uncaught, inside `frame()`, which has no try/catch and throws before its own `requestAnimationFrame(frame)` call — the overlay's rAF loop dies silently for the rest of the session (terminal itself is unaffected, separate callback).

Same class is reachable through the OSC channel, not just the URL: `pulse=toString` makes `PULSE_KINDS['toString']` truthy, and two frames later `pDef.c` / `pc[0]` throws the same way. This is one console command away in practice: `(pulse constructor)` from the dev console forwards the symbol straight through unvalidated (see the injection note below) and kills the overlay in one shot. `?shader=__proto__` is a quieter variant — `progCache['__proto__'] = null` actually nulls the object's own prototype rather than caching anything, confirmed in Node, though it doesn't throw on its own.

Fix is small: `Object.create(null)` for those three registries, or `Object.hasOwn(obj, name)` at every lookup site.

## Input-sanitization gap on the OSC emit path (console.lg -> shell)

`xsofy/console.lg`'s `emit-fx!` builds the OSC 7777 payload with plain string concatenation, and `hurt`/`pulse`'s args come straight from whatever the dev console's reader parses off `(hurt ...)`/`(pulse ...)` — no filtering of embedded control characters. A value containing a literal BEL terminates the (harmlessly-ignored-by-real-terminals) OSC 7777 sequence early, and anything after it in the string is emitted as ordinary terminal input — including a second, well-formed, *recognized* escape sequence of the typer's choosing (e.g. an OSC clipboard-write sequence on a terminal that supports it). The doc comment's safety claim ("real terminals discard unknown OSC ... output-only, so untouched") is true for a well-formed, unterminated-early sequence, but doesn't hold once the payload itself can smuggle a terminator plus a follow-on payload.

Practical severity today is low: the console is gated behind `dev?` (`xsofy/input.lg:44-51`, confirmed — the binding is only added in dev builds), so this is dev-only, self-inflicted tooling, not reachable by players. Flagging anyway since (a) it's the same crash surface as the JS bug above via a slightly different vector, and (b) the PR body notes "in-game pickup emits are a follow-up," which would mean less-trivially-controlled data (item/rune names) eventually feeding this same unsanitized sink. Worth a basic filter (strip control chars, or just validate `hurt` as a bounded int and restrict `pulse`'s `k` to the known `PULSE_KINDS` keys) at the `emit-fx!`/`emit-shell-fx!` boundary before that follow-up lands.

## Merge conflict (trivial, needs a rebase before this can move)

The only real conflict is in `xsofy/render.lg`: `main` has since added an `sfx/with-frame` wrapper at the same two spots in `render-full`/its incremental sibling where this PR inserts `(emit-shell-fx! world)`. Easy to resolve by composing the two (nest one inside the other) — not a design conflict, just needs the rebase done.

## Minor / latent, fine for a POC

- No re-entrancy guard on `startCrtOverlay()`'s rAF loop — called once today, but nothing stops a second call (e.g. a future context-loss-recovery path) from stacking two overlay canvases and two independent rAF loops. Cheap to guard with a `started` flag whenever this graduates.
- The shader-override toggle (`fxSignals.shader === v` -> clear, else set) only behaves correctly for one `shader=` key per OSC message; a hypothetical message with the key repeated would mis-toggle. Nothing in the current diff ever sends it twice, so this is latent, not live.

## Verified fine — no new issues found here

OSC parsing itself handles malformed input safely (missing `=`, empty segments, stray separators all degrade to ignored/zeroed fields, no throw, no unbounded growth — `fxSignals` only ever holds its fixed key set). `?renderer=dom` combined with `crt` correctly no-ops (`startCrtOverlay` bails when `glAddon` is null) rather than assuming a WebGL canvas exists. Resize/letterbox tracking only reallocates GPU state when sampled pixel dimensions actually change, not every frame. Shader program caching means runtime switching between the four shaders never recompiles and doesn't leak across a long session.

Didn't re-litigate what you already flagged yourself (`water.glsl` provenance, the per-frame `texImage2D` readback stall, the link-layer canvas-selection footgun) — agree those are real and worth resolving before this goes further, just nothing new to add on top of your own notes.

Net: fun concept, and the parts I could verify mechanically (parsing, resize, program lifecycle) hold up. The prototype-lookup crash is the one I'd actually fix regardless of whether this stays a spike — it's a one-line `Object.create(null)` change and it's currently one URL param or one dev command away from silently killing the entire effect for the rest of a session.

Issue #162, opened 2026-07-19
Burning while standing in water


nooga:
Standing in water while being shot by an imp does not prevent burning.

PR #170, opened 2026-07-22
tool: scenariotest: EDN scenario + play-loop harness

mparrett:
## Summary

- Adds `tools/scenariotest.lg` + `tools/scenariotest/build.lg`: load a handcrafted EDN arena (no dungeon gen) and drive it through the real `xsofy.play` / dispatch loop.
- Ships two visually readable starter scenarios: a 2×2 water pool for #162 and a 2/3/2 web patch for #161 / PR #166.
- Applies per-creature scenario patches to the exact entity created by `ent/spawn`, with regression coverage for overlapping positions.
- Sibling of `tools/lighttest.lg` (isolated TUI), but this one boots the game sim on purpose so turn-order / status / fire paths are faithful.

image

Screenshot 2026-07-22 at 9 43 55 AM

## Run

```bash
lg tools/scenariotest.lg                              # default: water-burning
lg tools/scenariotest.lg tools/scenarios/water-burning.edn
lg tools/scenariotest.lg tools/scenarios/web-trample.edn
lg tools/scenariotest.lg --check                      # build-only smoke, no TTY
```

In the TUI, normal game keys (`.` wait, movement keys, `Esc`/`Ctrl-C` quit). Unhandled menu screens are dismissed in this draft.

> **Web scenario visibility:** ash is tile 16, rendered as a dim gray `.`, while stone floor is a warmer brown `.`. The first trampled tile is hidden beneath `@` until the player moves away. For an unmistakable local comparison, temporarily change the scenario's `:fill` from `1` to `2`; ash `.` will then stand out against green grass `"`.

## Test plan

- [x] `lg tools/scenariotest.lg --check` prints the arena summary and exits 0
- [x] Both bundled scenario paths pass `--check`
- [x] Headless dispatched wait reproduces #162 on main (the water tile ignites and the player gains `:burning`)
- [x] Two up-left actions reproduce #161 on main (player moves through the open corner; center tile 19 remains) and demonstrate PR #166's fix (player stays; center tile becomes 16/ash)
- [x] New overlap regression verifies creature patches target the newly spawned entity
- [x] Existing suite: 300 tests, 2,782 assertions, 0 failures/errors
- [x] Interactive comparisons with the #162 and #166 fixes

Made with [Cursor](https://cursor.com)

### DIFF
--- a/tools/scenariotest/build.lg
+++ b/tools/scenariotest/build.lg
@@ -1,0 +1,130 @@
+(ns tools.scenariotest.build
+  "Build a playable xsofy world from a dev-scoped scenario EDN map.
+
+   Scenarios are handcrafted arenas — not dungeon gen. The shape is intentionally
+   small so a reviewer can stage one interaction (water+imp, web+struggle, …)
+   and drive it through the real play/dispatch loop.
+
+   Keys (all optional unless noted):
+     :name       string label
+     :desc       string logged on boot
+     :w :h       arena size (default 16×10)
+     :seed       gameplay seed (default 1)
+     :depth      floor depth (default 1)
+     :fill       interior tile id (default 1 = stone floor)
+     :paint      [[x y tile] ...] overlays after fill+border
+     :player     {:pos [x y] :statuses {...}}
+     :creatures  [{:species :fire-imp :pos [x y] :ai {...} :merge {...}} ...]
+     :gear?      true → give-starting-items (default false)
+     :log        initial message-log lines
+
+   The border is always stone walls (tile 8). Creatures are spawned via
+   ent/spawn then optionally :ai-merged / deep-merged so a scenario can wake
+   an AI or force a 100% fire-chance without forking the bestiary."
+  (:require [xsofy.terrain :as terrain]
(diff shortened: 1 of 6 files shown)


tools/scenariotest/build.lg
```diff
@@ -0,0 +1,131 @@
...
+                   (assoc :pos ppos)
+                   (cond-> (:statuses player-spec)
+                     (assoc :statuses (:statuses player-spec)))
+                   (cond-> (:body player-spec)
+                     (update :body merge (:body player-spec))))
+        boot-log (vec (concat
+                        [(str "scenariotest: " (or (:name sc) "scenario"))]
+                        (or (:log sc) [])
+                        (when (:desc sc) [(:desc sc)])))
+        base {:terrain terrain
```
nnunley, line 82: I wonder if this is an indication that we should have a struct or type defined to represent world state, since this sort of empty initialization should be already set up on the default base world state.

nnunley (approved):
It does look like there's some common code that should be extracted to avoid duplication and risks of having replicated state, particularly at the base of the world state initialization. That said, this is still non-blocking.

PR #171, opened 2026-07-22
fix(status): stop burning while standing in water

### LINKED ISSUE #162: Burning while standing in water

mparrett:
## What changed

- Prevent shallow and deep water tiles from being replaced by forced fire ignition.
- Route burning status application through a water-aware helper.
- Cover both water depths, normal floor behavior, and the fire-imp spell path with regression tests.

## Why

Fire runes applied `:burning` directly after impact, bypassing the movement-time water extinguish logic. As a result, a player already standing in water could catch fire when hit by a fire imp. Forced fire ignition could also replace the water tile and leave a stale fire TTL entry.

Players and creatures standing in shallow or deep water now remain unburned, while fire behavior on non-water terrain is unchanged.

Fixes #162.

## Validation

- `/Users/matt/projects-new/3p/let-go/lg xsofy/test/run.lg`
- 300 tests, 2,786 assertions, 0 failures, 0 errors
- `git diff --check`

Tested with the new harness in https://github.com/nooga/xsofy/pull/170

Screenshot 2026-07-22 at 10 33 02 AM

### DIFF
--- a/xsofy/fire.lg
+++ b/xsofy/fire.lg
@@ -18,6 +18,26 @@
 (def fire-spread-chance 30)  ;; percent per adjacent flammable per tick
 (def fire-default-ttl 5)
 (def protected-tiles #{13 14})  ;; stairs must survive dragon fire and rune fire
+(def water-tiles #{3 4})        ;; shallow + deep — extinguishes fire, never ignites
+
+(defn water-tile? [tile]
+  (contains? water-tiles tile))
+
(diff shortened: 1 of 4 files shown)


nnunley (approved):
I need to dig up my physics branch. I set up a multimap to indicate what should happen with different combined environment effects and a given cell

PR #173, opened 2026-07-23
feat(shell): mobile keypad rearrangement + font touch-ups

mparrett:
Polish for the existing Touch-UI web shell's on-screen keypad, sliced out of the shader demo spike (#168), since it's independent. The changes are sensible, unambiguous / pre-existing usability gaps, noticed during general play testing against the mobile shell. No renderer or shader code here, just `tools/xsofy-shell.html`. 

## Changes

- **PLAY pane — WAIT/ENTER swap.** Wait is the high-frequency key, so it moves up to the top row where enter was; enter takes wait's old corner. `y`/`n` also swap to `n`/`y` to match the landscape variant.
- **PLAY pane — `` ` `` (DEV) tile.** Fills the empty col-1/row-3 cell so the dev console is reachable on touch devices with no physical keyboard. The console is dev-gated game-side (`?mode=dev`), so the key is a no-op for regular players.
- **a–z pane — `(` and `)` keys.** Added left of `z`. Portrait: they replace the stagger spacer and backspace slims 3→2. Landscape: thin span-1 keys, esc/backspace slim to fit. Makes the dev console's command syntax typeable on touch.
- **Font default 14 → 18 (landscape/desktop).** 14 reads small on laptop screens too; portrait keeps its phone-tuned 22. A stored user choice (the font chip's localStorage value) still wins over both defaults, so nobody's existing setting changes.

image

image

image

## Notes

Every keypad row still tiles cleanly: PLAY stays 6×6→5 rows (30 cells), and the a–z rows sum to 20 units in both orientations. These exact hunks were exercised under Playwright as part of #168; the PR-preview smoke gate covers boot here.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/tools/xsofy-shell.html
+++ b/tools/xsofy-shell.html
@@ -714,7 +714,7 @@ body.dpad-cool #xsofy-shell button.dpad .label { color: rgba(168, 198, 224, 0.55
     { key: 'd', glyph: 'd', label: 'DROP' },    mv('y','↖','y'), mv('k','↑','k'), mv('u','↗','u'),
     { key: '<', glyph: '<', label: 'ASC' },     mv('h','←','h'), { key: 'x', glyph: 'x', label: 'AUTO' }, mv('l','→','l'),
     { key: '>', glyph: '>', label: 'DESC' },    mv('b','↙','b'), mv('j','↓','j'), mv('n','↘','n'),
-    { key: 'r', glyph: 'r', label: 'RUNES' },   { key: 'y', glyph: 'y', label: 'YES' },   { key: 'n', glyph: 'n', label: 'NO' },    { key: '.', glyph: '·', label: 'WAIT', repeat: true },
+    { key: 'r', glyph: 'r', label: 'RUNES' },   { key: 'n', glyph: 'n', label: 'NO' },    { key: 'y', glyph: 'y', label: 'YES' },   { key: '.', glyph: '·', label: 'WAIT', repeat: true },
   ]};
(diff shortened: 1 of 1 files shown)


mparrett:

nnunley (approved):
I'd would be nice to have more on screen real-estate for display.

We should do a design spike for constrained screens . Maybe something that opens radial, context-sensitive menus on long press, move to position/attack on tap, etc. Alternative, an on screen controller overlay with directions, menu context menus, etc.

mparrett:
> I'd would be nice to have more on screen real-estate for display.
> 
> We should do a design spike for constrained screens . Maybe something that opens radial, context-sensitive menus on long press, move to position/attack on tap, etc. Alternative, an on screen controller overlay with directions, menu context menus, etc.

@nnunley Valid point and agree we should spike on it. Taking this as non-blocking given the approval. Thanks for taking a look and endorsing. For initial gameplay, the small changes here remove substantial awkwardness.

**Some disclosure/context for the current design that this PR makes adjustments to:** The existing TUI works well enough in a square terminal area. Taller aspect ratios also work, but needed a closer look for some cases. I ran with a square-ish constraint for the terminal area + remainder as input as the straightforward first cut. The current UI mainly grew around these practical constraints, rather than being defined by a proper spike on design options. That said, I did play test this quite a bit and it's surprisingly workable for what it is.

mparrett:
@nooga I'm going to run with this PR given that it is approved and unblocks the current mobile gameplay, but please know know I'd still value your input. I understand you have a ton on your plate right now.

PR #176, opened 2026-07-24
refactor(world): `empty-world` — one canonical world-state skeleton

### LINKED ISSUE #172: No canonical empty/skeleton world-state constructor — two hand-rolled copies already disagree
Prompted by [@nnunley's review comment on #170](https://github.com/nooga/xsofy/pull/170#discussion_r3634333246), on the `base` world map in `tools/scenariotest/build.lg`:

> I wonder if this is an indication that we should have a struct or type defined to represent world state, since this sort of empty initialization should be already set up on the default base world state.

And restated in [the approval review](https://github.com/nooga/xsofy/pull/170#pullrequestreview-4759381109):

> It does look like there's some common code that should be extracted to avoid duplication and risks of having replicated state, particularly at the base of the world state initialization. That said, this is still non-blocking.

`xsofy.world/make-world` is the only real world constructor, and it's fused to `generate-floor` — full dungeon gen. Anything that wants a world without dungeon gen (a property-test generator, a scenario harness) has no lighter-weight constructor to call, so it hand-rolls the map literal instead. Two examples already exist, and they disagree:

- `xsofy.check_gen/minimal-stone-room` (property-test generator) sets **7 keys**: `:terrain :width :height :entities :turn :gold :rune-table`.
- `tools/scenariotest/build.lg`'s `base` (added in #170) sets **19 keys**, including `:fov :memory :lights :stains :gases :fire-ttl :depth :floors :running :seed :seed-input :action-log` — none of which `minimal-stone-room` sets.

Nothing enforces that either literal tracks what `update-world` and friends actually read. If the world map grows a key that some code path depends on, `minimal-stone-room`-derived property tests silently keep running against a world missing that key, covering less than they appear to. `gen-rich-world` already patches `:depth`, `:floors`, and `:fire-ttl` onto `gen-minimal-world`'s shape — the in-code comment says `:depth`/`:floors` are "needed by save/restore code paths"; `:fire-ttl` gets added the same way, unexplained, for the fire properties that need it.

### Proposal

Pull a single default/empty world-state constructor out of `xsofy.world`, independent of `generate-floor`, that both the property-test generator and the scenario harness build on:

```clojure
(defn empty-world
  "Bare world skeleton: no floor, no entities but the player. Not playable
   until terrain/entities are populated by a caller."
  [w h seed]
  {:terrain nil :width w :height h
   :entities {} :fov #{} :memory {} :lights nil
   :stains {} :gases {} :fire-ttl {}
   :log [] :turn 0 :depth 1 :floors {} :gold 0 :rune-table {}
   :running true :seed seed :seed-input seed :action-log []})
```

`make-world` becomes `(-> (empty-world w h seed) (assoc :terrain ...) generate-floor ...)`, `minimal-stone-room` and `scenariotest/build.lg`'s `base` both start from `empty-world` instead of hand-listing keys, and a future field addition is one edit instead of an audit across three files.

### Open questions

- Should this be a plain map constructor (as sketched) or a `defrecord`/spec that can validate shape at construction time? A record would catch the "missing key" class of bug harder than a doc comment does, at the cost of `deftype`/protocol churn across the codebase that treats world as an ordinary map.
- Where does `:terrain` belong in the skeleton — `nil` (caller must fill it before the world is usable) or a zero-size/degenerate terrain so `empty-world` alone never crashes?

#170's `base` map is a reasonable rendering of the current shape — one more copy of it. Happy to take a pass at extracting `empty-world` if there's appetite.

mparrett:
# refactor(world): `empty-world` — one canonical world-state skeleton

Closes #172. Follows up @nnunley's review comments on #170.

## Why

`xsofy.world/make-world` is the only world constructor and it's fused to `generate-floor`, so it always runs full dungeon gen. Anything that wants a world without dungeon gen has no lighter constructor to call and hand-rolls the map literal instead. Four copies existed, in two shapes that disagree:

| Site | Keys set |
|---|---|
| `check-gen/minimal-stone-room` | 7 |
| `check-gen/build-feature-room` | 7 |
| `combat_prop`'s `fresh` fixture | 7 |
| `tools/scenariotest/build.lg`'s `base` (added in #170) | 20 |

The three 7-key literals have identical key sets; `base` sets 20. So the split is the one #172 named — a test-side shape and a scenario-side shape, 13 keys apart.

Nothing tied any of them to what `update-world` and friends read. Add a field to the world map and the generated worlds silently fall behind: a property test keeps running green against a world missing the very key the code under test reads, covering less than it looks like it covers. `gen-rich-world` was already patching `:depth`, `:floors`, and `:fire-ttl` back on after the fact to close part of that gap.

## What

`world/empty-world` is now the one place the construction-time world shape is written down. `generate-floor` builds on it, and each hand-rolled literal starts from it and assocs only what's specific to that caller:

```clojure
(assoc (world/empty-world w h seed)
       :terrain terrain
       :entities {:player player}
       :log boot-log
       :depth depth)
```

Adding a *construction-time* field is one edit here instead of an audit across four files. A field that also has to survive a floor transition still needs `restore-floor` and `change-floor` updated by hand — see the follow-up note below.

## Decisions

**Plain map, not a `defrecord`.** #172 raised the option. A record would catch the missing-key class harder than a docstring does, but the codebase treats world as an ordinary map throughout: `assoc`, `get-in`, `:keys` destructuring, `serialize/canonical-bytes` walking it generically. The churn isn't worth it for a shape one function now owns. Happy to run with this or switch to a record if you'd rather have construction-time validation.

**`:terrain` is nil, not a degenerate grid.** The other open question in #172. A zero-size grid would let `empty-world` alone avoid crashing, at the price of a world whose `:width`/`:height` disagree with its terrain. Nil means a caller that forgets to fill it fails at the first `tget` instead of reading plausible garbage.

**The skeleton covers construction-time shape only.** `:rooms-cache`, `:next-id`, and `:scheme` belong to the systems that add them later, so they stay out.

**`invariant-test`'s sparse room is unchanged.** The tests for `world-invariants?` shouldn't be built from the production constructor, or a broken `empty-world` goes invisible to exactly the predicate meant to catch it.

**`generate-floor` merge order.** The skeleton goes down first, then `world-after-terrain` merges over it so terrain gen's advanced `:seed` lineage still wins. The skeleton contributes `:seed-input` as the origin seed, which is what that key means on both call paths. Both callers reset `:seed` afterward, same as before.

## Verification

World content is unchanged except for one added empty key. `make-world` at seeds 1, 42, and 1000, plus a d1→d2→d3→d2 floor-transition chain, produce the same `canonical-bytes` hash on this branch as on `main` once the newly-present `:fire-ttl {}` is removed. That key is new on every freshly generated world, so a whole-world `checkpoint-hash` does change; nothing persists or asserts against one (replay codes are `(seed, action-log)`, `dag/commit-id` is `:seed` alone).

- `make test` — 307 tests, 2803 assertions, 0 fail
- `make smoke-lg` — OK
- `lg tools/scenariotest.lg --check` on both scenarios — OK

### DIFF
--- a/tools/scenariotest/build.lg
+++ b/tools/scenariotest/build.lg
@@ -79,26 +79,11 @@
                         [(str "scenariotest: " (or (:name sc) "scenario"))]
                         (or (:log sc) [])
                         (when (:desc sc) [(:desc sc)])))
-        base {:terrain terrain
-              :width w
-              :height h
-              :entities {:player player}
-              :fov #{}
-              :memory {}
-              :lights nil
-              :stains {}
-              :gases {}
-              :fire-ttl {}
-              :log boot-log
-              :turn 0
-              :depth depth
-              :floors {}
-              :gold 0
-              :rune-table {}
-              :running true
-              :seed seed
-              :seed-input seed
-              :action-log []}
+        base (assoc (world/empty-world w h seed)
+                    :terrain terrain
+                    :entities {:player player}
+                    :log boot-log
+                    :depth depth)
         w0 (reduce spawn-creature base creatures)
         w1 (if gear? (world/give-starting-items w0) w0)]
     (world/update-fov w1)))
--- a/xsofy/check_gen.lg
+++ b/xsofy/check_gen.lg
@@ -29,6 +29,7 @@
   (:require [xsofy.check :as c]
             [xsofy.terrain :as terrain]
             [xsofy.entities :as ent]
+            [xsofy.world :as world]
             ;; Load template registries (defcreature / defitem populate
             ;; the shared @ent/templates atom).
             [xsofy.bestiary]
@@ -41,19 +42,19 @@
 ;; --- Minimal world for spell/world properties ---
 
 (defn minimal-stone-room
-  "Build a 10x10 stone-floor room with stone walls. No entities."
+  "Build a 10x10 stone-floor room with stone walls. No entities.
+
+   Shape comes from world/empty-world so these generated worlds stay
+   structurally identical to the ones the game builds — a property test
+   run against a world missing a key the code under test reads covers
+   less than it appears to. Seed 0 matches what det/advance already
+   defaults to when :seed is absent, so rolls are unchanged."
   []
   (let [w 10
         h 10
         t (terrain/make-terrain w h)]
     (terrain/carve-room! t w 1 1 (- w 2) (- h 2))
-    {:terrain t
-     :width w
-     :height h
-     :entities {}
-     :turn 0
-     :gold 0
-     :rune-table {}}))
+    (assoc (world/empty-world w h 0) :terrain t)))
 
 ;; --- Player ---
 
@@ -149,8 +150,7 @@
     (loop [i 0
            fires []]
       (if (= i (* inner-w inner-h))
-        {:world {:terrain t :width w :height h
-                 :entities {} :turn 0 :gold 0 :rune-table {}}
+        {:world (assoc (world/empty-world w h 0) :terrain t)
          :fire-cells fires}
         (let [x (+ 1 (mod i inner-w))
               y (+ 1 (quot i inner-w))
@@ -251,8 +251,9 @@
         (assoc-in [:entities :player] player)
         (spawn-creatures creature-spawns)
         (spawn-items item-spawns)
-        ;; depth + floors map needed by save/restore code paths
-        (assoc :depth 1 :floors {} :fire-ttl fire-ttls))))
+        ;; :depth/:floors come from the skeleton; :fire-ttl is real
+        ;; generated data, one entry per fire tile in the room.
+        (assoc :fire-ttl fire-ttls))))
 
 ;; --- Picking two distinct entities from a world (for combat properties) ---
 ;; A combat property needs an attacker-id and a target-id. We pick the
--- a/xsofy/test/combat_prop.lg
+++ b/xsofy/test/combat_prop.lg
@@ -118,8 +118,8 @@
   [world g/gen-rich-world]
   (let [[player items] (w/extract-player-items world)
         ;; inject into a fresh empty world at original pos
-        fresh {:entities {} :turn 0 :width (:width world) :height (:height world)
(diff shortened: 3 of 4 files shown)


nnunley (approved):

mparrett:
A second review pass over this turned up things worth recording before merge. None of them change the diff as approved; they are all follow-up work.

**`percept.lg:63` becomes a latent trap.** `entity-fov` uses key *presence* as its "player FOV already computed" sentinel:

```clojure
(and (= entity-id :player) (some? (:fov world))) (:fov world)
```

The test-side worlds previously had no `:fov` key, so they fell through and computed fresh FOV. They now get `:fov #{}` from the skeleton, which `some?` accepts, so the empty set is returned as though it were a result. Nothing calls it today — `percept` is reached only from `balance.lg`, which runs on real `make-world` output — so this is not a live regression, but it will surprise the first percept-based property test. `fov/compute-fov` seeds its result with `#{[ox oy]}` (`fov.lg:80`), so a computed FOV is never empty and `(seq (:fov world))` is a sound replacement.

**The docstring overclaims.** `world.lg:195` says "this is the one place that shape is written down." `restore-floor` (`world.lg:400-421`) is a 21-key world literal 180 lines below it, and `change-floor` (`world.lg:694-702`) carries eight more keys by hand. For any field that has to survive a floor transition, the PR body's "one edit instead of an audit" is wrong — it is three. Correcting the docstring and folding both into the skeleton is the follow-up.

**`:next-id` is excluded for the wrong stated reason.** The docstring attributes it to entity spawning, but `entities/next-id` is seed-derived and never writes the key; the only writer is `det/next-id`, called only from `det_test.lg`. Right to leave it out, wrong justification.

**Corrected the key counts in the body.** The three test-side literals set 7 keys each with identical key sets, and scenariotest's `base` set 20 — the same 20 the skeleton has. So there were two distinct shapes, not four disagreeing ones, which is what #172 said in the first place. The original table overstated it.

Filed the `:title`/`:quest-item` transition drift as #177 — it is pre-existing and independent of this change.

PR #178, opened 2026-07-24
fix(world): carry run-scoped fields across floor transitions

### LINKED ISSUE #177: World fields set after make-world are dropped on floor transition (:title, :quest-item)
`main.lg:336-338` assocs three fields onto the world immediately after `make-world`:

```clojure
(-> (world/make-world 79 30 seed)
    (assoc :title title
           :scheme scheme
           :quest-item (:item quest)))
```

Descending drops two of the three. `change-floor`'s regen branch carries `:scheme` by hand (`world.lg:700`), and `restore-floor`'s world literal carries `:scheme` by hand (`world.lg:416`). Neither carries `:title` or `:quest-item`, and nothing else puts them back.

Probe against a real world (`make-world` 40x20 seed 7, then descend and return):

```
d1          title= "The Chasms"  scheme= [1 2 3]  quest= :amulet  rooms-cache? true
d2-new      title= nil           scheme= [1 2 3]  quest= nil      rooms-cache? true
d1-revisit  title= nil           scheme= [1 2 3]  quest= nil      rooms-cache? false
```

## Impact

`render.lg:1196` reads `(or (:title world) "")` for the death-screen tombstone, so a player who dies below depth 1 gets a tombstone with no game title. That is the live symptom.

`:quest-item` has no reader today — `ui_bridge.lg:44` emits the quest at startup from its own arguments, not from the world — so its loss is latent. It is the one that will bite: the next feature to read `:quest-item` mid-run will read it as nil on every floor but the first.

`restore-floor` also drops `:rooms-cache` (the probe's third column), which makes `spawn-runestones` (`world.lg:1061`) no-op on a revisited floor. That may well be intended, since the cache is per-floor derived data, but it is unstated.

## Root cause

Three separate places declare what a world holds: `world/empty-world`, `restore-floor`'s 21-key literal (`world.lg:400-421`), and `change-floor`'s carry-over assoc (`world.lg:694-702`). #176 unified *construction* onto `empty-world`; the transition path still spells the shape out twice by hand. A field only survives a floor change if someone remembers to add it in two more places, and `:title`/`:quest-item` are the proof that nobody did.

This is the drift #172 describes, with a symptom attached.

## Proposal

Name the distinction the code is actually making — fields rebuilt per floor versus fields that belong to the run — as an explicit set that both `change-floor` and `restore-floor` consume, instead of two hand-maintained literals. `:title`, `:scheme`, and `:quest-item` are run-scoped; `:terrain`, `:entities`, `:lights`, and friends are floor-scoped.

Happy to take this one as the follow-up to #176.

mparrett:
# fix(world): carry run-scoped fields across floor transitions

Closes #177. Follow-up to #176.

## The bug

`main.lg:336-338` assocs `:title`, `:scheme`, and `:quest-item` onto the world right after `make-world`. Descending drops two of the three:

```
d1          title= "The Chasms"  scheme= [1 2 3]  quest= :amulet
d2-new      title= nil           scheme= [1 2 3]  quest= nil
d1-revisit  title= nil           scheme= [1 2 3]  quest= nil
```

`render.lg:1196` reads `(or (:title world) "")` for the death-screen tombstone, so a player who dies below depth 1 gets one with no game title. `:quest-item` has no reader yet, so its loss was silent and waiting for the first feature to want it.

The cause is that two places hand-listed what survives a transition — `change-floor`'s carry-over assoc and `restore-floor`'s 21-key world literal — and only `:scheme` made it onto both lists.

## The fix

#176 unified world *construction* onto `empty-world` and left the transition path
spelling the shape out by hand, which is exactly how a field ends up construction-correct and transition-lossy at the same time. This finishes that job:

- `run-scoped-keys` names what belongs to the run rather than the floor.
`carry-run-state` is the single thing both transition paths call.
- `restore-floor` builds on `empty-world` like every other constructor, assoc'ing
only the floor-scoped fields it reads back out of `:floors`.
- `empty-world`'s docstring stops claiming to be the only place the shape is
written down — `restore-floor` sat 180 lines below it proving otherwise. It now names `run-scoped-keys` as the companion answer to "does this field survive a descent?", since a new world field needs an answer in both places.

## Two details worth a reviewer's eye

**`:seed` is deliberately excluded from `run-scoped-keys`.** It heads the RNG chain, and both callers restore it as their *final* step so that `spawn-runestones` in between consumes the floor lineage rather than the gameplay one. Carrying it with the rest would move that restore earlier and shift every subsequent roll. The exclusion is load-bearing, not an oversight, and it's documented at the definition.

**`generate-floor` takes `:seed` across explicitly** instead of merging `world-after-terrain` wholesale. Behavior is identical today — terrain gen threads nothing but `:seed` — but the blanket merge granted it precedence over every skeleton key, so the day it starts threading more state the skeleton would lose silently.

Also corrected in passing: `empty-world`'s docstring claimed `:next-id` comes from entity spawning. `entities/next-id` is seed-derived and never writes that key; the only writer is `det/next-id`, called only from tests.

## Verification

`world_shape_test.lg` pins both halves of the shape: every skeleton key present after construction, after a descent, and after a revisit; and the run-scoped fields surviving both. Its four `:title`/`:quest-item` assertions fail on `main` and pass here. The three skeleton-key tests pass on `main` too — they are regression guards for the next field, not evidence of a second bug.

Determinism is unchanged. A d1→d2→d3→d2→d1 chain produces identical `canonical-bytes` hashes and identical `:seed` at every step before and after, once the newly-surviving `:title`/`:quest-item` are excluded from the comparison.

- `make test` — 314 tests, 2818 assertions, 0 fail
- `make smoke-lg` — OK
- `lg tools/scenariotest.lg --check` on both scenarios — OK

One thing this PR does not touch: `percept.lg:63` uses key presence as its "player FOV already computed" sentinel, which #176 made misleading for skeleton-built worlds. It's a different concern in a different file and gets its own PR.

### DIFF
--- a/xsofy/test/run.lg
+++ b/xsofy/test/run.lg
@@ -10,6 +10,7 @@
             [xsofy.test.hash-test]
             [xsofy.test.det-test]
             [xsofy.test.world-seed-test]
+            [xsofy.test.world-shape-test]
             [xsofy.test.dispatch-test]
             [xsofy.test.serialize-test]
             [xsofy.test.dag-test]
--- a/xsofy/test/world_shape_test.lg
+++ b/xsofy/test/world_shape_test.lg
@@ -1,0 +1,72 @@
+(ns xsofy.test.world-shape-test
+  "Pins the world map's shape across construction and floor transitions.
+
+   world/empty-world declares what a world holds; world/run-scoped-keys
+   declares what survives a descent. Neither is enforced by anything the
+   compiler can see, so these tests are the enforcement: a field added to
+   the skeleton and forgotten in the transition path fails here rather
+   than three months later as a nil on depth 2 (#177)."
+  (:require [test :refer [deftest is testing]]
+            [xsofy.world :as w]))
+
+(def ^:private skeleton-keys
+  (set (keys (w/empty-world 10 10 0))))
+
+(defn- missing-keys
+  "Skeleton keys absent from world. Absent, not nil-valued: :terrain and
+   :lights are legitimately nil on a live world."
+  [world]
+  (vec (sort (filter (fn [k] (not (contains? world k))) skeleton-keys))))
+
+;; A small world keeps floor generation cheap; the shape doesn't depend on size.
+(defn- run-world []
+  (-> (w/make-world 40 20 7)
+      (assoc :title "The Chasms"
+             :scheme [1 2 3]
+             :quest-item :amulet)))
+
+(deftest make-world-carries-the-whole-skeleton
+  (testing "a constructed world holds every key empty-world declares"
+    (is (= [] (missing-keys (w/make-world 40 20 7))))))
+
+(deftest descending-carries-the-whole-skeleton
+  (testing "a newly generated floor holds every key empty-world declares"
+    (is (= [] (missing-keys (w/change-floor (run-world) 2))))))
+
+(deftest revisiting-carries-the-whole-skeleton
+  (testing "a restored floor holds every key empty-world declares"
+    (let [d2 (w/change-floor (run-world) 2)]
+      (is (= [] (missing-keys (w/change-floor d2 1)))))))
+
+;; --- Run-scoped fields survive a transition (#177) ---
+
+(deftest run-scoped-fields-survive-descending
+  (testing ":title, :scheme and :quest-item outlive the floor they were set on"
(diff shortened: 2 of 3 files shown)


mparrett:

nnunley (approved):

PR #179, opened 2026-07-24
fix(percept): treat an empty :fov as uncomputed, not as a result

mparrett:
Follow-up to #176. Independent of #177 — different file, different concern.

## The problem

`entity-fov` decides whether the player's FOV is already cached by asking whether the key is there:

```clojure
(and (= entity-id :player) (some? (:fov world))) (:fov world)
```

That reads the world's *shape* as a signal, and it held only while worlds that hadn't been through `update-fov` had no `:fov` key at all. #176 ended that: `empty-world` seeds `:fov` to `#{}`, so a skeleton-built world carries the key with nothing in it. `entity-fov` now hands that empty set back as though it were a computed result, and every "can the player see X" question answers no.

Nothing reaches it today. `percept` is required only by `balance.lg`, which runs on `make-world` output that `generate-floor` has already FOV'd. The cost lands on whoever writes the first percept-based property test against `check-gen`'s generators and spends an afternoon on visibility assertions that quietly answer false.

## The fix

`(seq (:fov world))`. `fov/compute-fov` seeds its result with the origin cell (`fov.lg:80`), so a computed FOV always holds at least the entity's own position, and an empty one can only mean uncomputed. That makes emptiness a sound sentinel where presence no longer is.

The alternative — dropping `:fov` from `empty-world` to restore the old absent-key signal — would reintroduce exactly the shape drift #172 was about, and leave the next reader of the skeleton wondering why one key is missing.

## Verification

New `percept_test.lg` covers both sides of the branch, the real `update-fov` path, and the NPC path that must never read the player's cache. Its two player-side assertions fail against the current `percept.lg` and pass with this change.

- `make test` — 311 tests, 2810 assertions, 0 fail

### DIFF
--- a/xsofy/test/percept_test.lg
+++ b/xsofy/test/percept_test.lg
@@ -1,0 +1,51 @@
+(ns xsofy.test.percept-test
+  "entity-fov's cache sentinel.
+
+   The player's FOV is precomputed on the world by update-fov; NPCs get it
+   computed on demand. entity-fov decides which case it's in by looking at
+   :fov, so what an empty :fov means is load-bearing — and since #176 gave
+   empty-world a :fov of #{}, a world that hasn't been through update-fov
+   carries the key with nothing in it."
+  (:require [test :refer [deftest is testing]]
+            [xsofy.percept :as percept]
+            [xsofy.terrain :as terrain]
+            [xsofy.world :as w]))
+
+(defn- room-world
+  "A 10x10 stone room with the player at [2 2] and no FOV computed yet —
+   the shape a property-test generator or scenario builder produces."
+  []
+  (let [t (terrain/make-terrain 10 10)]
+    (terrain/carve-room! t 10 1 1 8 8)
+    (-> (w/empty-world 10 10 0)
+        (assoc :terrain t)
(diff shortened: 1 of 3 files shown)


mparrett:

xsofy/test/percept_test.lg
nnunley, line 1: Consider adding an obstruction in the middle of the room to validate that objects/mobs on the other side of the obstruction can't see each other via fov.

nnunley (approved):

xsofy/test/percept_test.lg
mparrett: Agreed, and it's a real gap: nothing under `xsofy/test/` references `compute-fov`, `has-clear-line?`, or `transparent?` today, so a `compute-fov` that ignored walls entirely would pass the suite.

Filed as #184 rather than holding this PR, since the suggestion came in with your approval and covers existing FOV behavior rather than what this change touches. The issue sketches the room-plus-wall-segment shape, including an unobstructed control so the test can't pass by returning nothing, and notes `has-clear-line?` needs the same coverage.

PR #180, opened 2026-07-25
Top-down perf: cull never-seen cells (22x), plus frame timing and two input fixes

mparrett:
Four commits on top of `gl-frontend` at `cc43ffb`, so this merges with a button. Nothing is rebased and nothing of yours moves.

The presentation state machine and the fall into first person are a clever way to get a 3D view out of a glyph grid without giving up the grid, and porting the frame loop onto `ui/run-loop` made the render path easy to follow from the outside. Everything below is downstream of that.

We were rebasing the branch onto current `main` for a downstream experiment and profiled it along the way. The results seemed worth handing back.

## What's here

**Frame timing (`XSOFY_GL_PERF=N`).** The branch had no frame instrumentation, so its cost was unmeasured rather than unoptimized. This reports two clocks every N frames:

- `cpu` — `BeginFrame` to just before `EndFrame`, the part we control.
- `interval` — render entry to render entry, including the vsync block
inside `EndFrame`.

The second clock is what made the diagnosis possible. A vsync-bound loop shows `interval` pinned near 16.7ms while `cpu` moves freely. Here the two tracked each other, which pointed at our own vertex building.

**Culling never-seen cells, which is the win.** A cell that is neither in FOV nor remembered draws a space glyph on a black background, so it contributes nothing visible, but it still costs a map in `grid-cells` plus a background quad and a glyph quad in `grid-verts`. At seed 42 that was 2,330 of 2,370 cells.

| | before | after |
|---|---|---|
| cells | 2,370 | 40 |
| verts per batch | 14,220 | 240 |
| cpu per frame | ~400ms | ~18ms |

The captured frame is byte-identical before and after, which is the property that makes this safe: a 22x reduction that changes no pixels.

Culling is opt-in, and off in first person by design. There the unseen floor quads are the black ground plane under the horizon and the clear colour reads as sky, so dropping them would show sky through the floor. It is sound in top-down because top-down clears to black, and the cells being dropped are what was painting that black.

One caveat on the absolute numbers: ~18ms was measured on a quiet machine, and the same build measures around 55ms under heavy load. The ratio was taken back-to-back in a single machine state, so it holds; the frames-per-second figure is load-dependent and should be read as indicative.

**A `dissoc` hoist in `grid-verts`,** which is bookkeeping rather than a win. It rebuilt the options map once per cell. After culling it buys nothing measurable at 40 cells, and it is here only because the first-person path still sends all 2,370 cells through `grid-verts`.

**Two input fixes,** both present since the branch first became playable:

- *Every keypress moved two tiles.* GLFW fires the key callback and the
char callback for one physical press, so `k` arrived as both `key:k` and `char:k`, and both parsed to `:up` and both dispatched.
- *The arrow keys did nothing at all.* `input/parse-key` compares whole
strings against escape-prefixed bindings, and the mapping emitted a bare two-character sequence, which matched nothing and parsed to `:unknown`. That is why the only usable movement keys were the ones that double-fired.

Keeping the `char:` event resolves both, because it alone carries shift state, and xsofy binds shifted keys that `key:` reports only by their lowercase name (`K`/`J`/`H`/`L` to run, `>` to descend, `?` for help). Before this, Shift+K dispatched `:up` and `:run-up` together. Verified per key class against the real parser: `k` gives one `:up`, Shift+K gives `:run-up` only, Up arrow gives `:up`, `>` gives `:descend`, and a held arrow repeats.

**Taking only part of it.** The `dissoc` hoist touches `grid.lg` alone and is independent. The other three all touch `main_gl.lg`, so they cherry-pick cleanly in order but not individually.

## Two rebase findings

Neither is in this PR, but both come up whenever the branch moves forward.

**The test suite hangs on this branch's current base.** It reaches `decode . encode = identity on (seed, actions)` in `replay-test` in about 70 seconds, then spins and allocates without bound. We measured 12.4 GB resident before killing it. Nothing in this PR causes it: clean `cc43ffb` with none of our commits reproduces it exactly, and none of these commits touch replay. On current `main` the same suite finishes in 60 seconds at 444 tests and 3,066 assertions with nothing failing, so it appears to be fixed upstream already.

**`title.lg` conflicts on rebase, and the natural resolution is wrong.** Taking the branch side drops the two `sfx/with-frame` calls that arrived with #174. The resolution that preserves both keeps main's body verbatim and grafts only the guard onto it. One detail that is easy to miss: `os/getenv` returns `""` rather than `nil` for an unset variable, so the guard has to test by count.

## Not done, on purpose

- **Caching static terrain geometry.** This was our next planned step
until culling obsoleted it. It assumed rebuilding 2,370 cells per frame; the real number is now 40.
- **Driving the loop from FOV plus memory** instead of scanning the grid.
We tried it and it came out 2.5x slower, because building the position set each frame costs more than the tight scan does. Reverted, and noted here so nobody re-derives it.
- **Escape still does not quit.** It maps to `""` and parses to
`:unknown`. We took that for intentional and left it.
- **Shift plus arrow cannot run.** glplat's key callback accepts GLFW's
modifier argument and discards it, so this needs a change on the let-go side rather than here.

## Where the remaining time goes

Of the ~18ms: `grid-cells` is roughly 2 to 4ms, `grid-verts` roughly 5 to 11ms, and the rest is HUD, VFX, and submit. At about 175 microseconds per cell, `grid-verts` is no longer doing algorithmic work that can be deleted; that is interpretation overhead on flat float-vector construction, which makes the next lever a let-go one rather than a game one.

Merge as-is if it looks right, or take the input fixes alone and leave the perf work. The one piece I would not defer is the rebase: the suite hang makes the branch hard to validate as it stands, and it is already fixed on `main`.

### DIFF
--- a/main_gl.lg
+++ b/main_gl.lg
@@ -123,20 +123,44 @@
 ;; Input Parsing
 ;; ==========================================================
 
+(defn arrow-seq
+  "The ANSI sequence input/parse-key binds for an arrow key name, else nil."
+  [name]
+  (case name
+    "up" "\u001b[A"
+    "down" "\u001b[B"
+    "left" "\u001b[D"
+    "right" "\u001b[C"
+    nil))
+
 (defn parse-glplat-event
-  "Parse a glplat event string (e.g., 'key:w', 'key:escape', 'char:x') to a key string
-   that input/parse-key can consume."
+  "Parse a glplat event string (e.g., 'key:w', 'key:escape', 'char:x') to a key
+   string that input/parse-key can consume. nil means ignore this event.
+
+   Two things this has to get right, and originally got neither:
+
+   Arrows need the ESC prefix. input/parse-key compares whole strings against
+   bindings like \"\\u001b[A\", so a bare \"[A\" matched nothing and the arrow
+   keys did not move the player at all.
+
+   A printable key arrives twice. GLFW fires the key callback (key:k) and the
+   char callback (char:k) for one physical press, and dispatching both moved
+   the player two tiles per keystroke. The char: event is the one to keep — it
+   is the only one carrying shift state, and xsofy binds shifted keys (K/J/H/L
+   run, > descend, ? help) that key: reports only as their lowercase name. So a
+   key:/key-repeat: event naming a single printable character is dropped here
+   and its char: twin does the work."
   [event-str]
   (cond
-    (string/starts-with? event-str "key:escape") ""
-    (string/starts-with? event-str "key:") (let [key (subs event-str 4)]
-                                             (case key
-                                               "up" "[A"
-                                               "down" "[B"
-                                               "left" "[D"
-                                               "right" "[C"
-                                               key))  ;; fallback to raw key
-    (string/starts-with? event-str "key-repeat:") (subs event-str 11)
+    (string/starts-with? event-str "key:escape") ""
+    (string/starts-with? event-str "key:")
+    (let [name (subs event-str 4)]
+      (or (arrow-seq name)
+          (when (> (count name) 1) name)))
+    (string/starts-with? event-str "key-repeat:")
+    (let [name (subs event-str 11)]
+      (or (arrow-seq name)
+          (when (> (count name) 1) name)))
     (string/starts-with? event-str "char:") (subs event-str 5)
     :else nil))
 
--- a/main_gl.lg
+++ b/main_gl.lg
@@ -160,6 +160,58 @@
         (#(reduce world/log-msg % (:messages quest))))))
 
 ;; ==========================================================
+;; Frame-time instrumentation (XSOFY_GL_PERF)
+;; ==========================================================
+
+;; Two clocks, because they answer different questions and only one of them
+;; responds to render work. `cpu` is BeginFrame → just-before-EndFrame: the
+;; segment we actually control (bridge → grid-verts → SubmitTriangles).
+;; `interval` is render-entry to render-entry, which includes EndFrame's vsync
+;; block. A change that halves vertex work moves `cpu` and leaves `interval`
+;; pinned near the refresh period — so reading only wall-clock frame time would
+;; report every optimization as a no-op.
+
+(defn new-perf-acc []
+  {:n 0 :cpu-sum 0.0 :cpu-min 1.0e9 :cpu-max 0.0
+   :int-sum 0.0 :last-t nil :bg 0 :glyphs 0 :total 0})
+
+(defn perf-note-verts!
+  "Record the batch sizes the current frame emitted, so a culling change shows
+   up as vertices, not only as milliseconds."
+  [perf* bg glyphs]
+  (when perf*
+    (swap! perf* assoc :bg bg :glyphs glyphs)))
+
+(defn perf-record!
+  "Fold one frame's timings in; emit a line every `every` frames and reset."
+  [perf* every t-entry cpu-ms]
+  (when perf*
+    (let [a (swap! perf* (fn [a]
+                           (-> a
+                               (assoc :n (inc (:n a))
+                                      :total (inc (:total a))
+                                      :last-t t-entry)
+                               (assoc :cpu-sum (+ (:cpu-sum a) cpu-ms))
+                               (assoc :cpu-min (min (:cpu-min a) cpu-ms))
+                               (assoc :cpu-max (max (:cpu-max a) cpu-ms))
+                               (assoc :int-sum (+ (:int-sum a)
+                                                  (if (:last-t a)
+                                                    (* 1000.0 (- t-entry (:last-t a)))
+                                                    0.0))))))]
+      (when (>= (:n a) every)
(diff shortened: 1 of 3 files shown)


mparrett:
CI is red here, and it is the base rather than this PR.

The `latest (upstream smoke test)` leg hangs in `play_test.lg` at `play-step-dispatches-explicit-action`, producing no output for nine minutes before the runner kills it. That test requires `xsofy.world`, `xsofy.dispatch` and `xsofy.play`. This PR touches `main_gl.lg`, `gl/bridge.lg` and `gl/grid.lg`, changes no test file, and the hanging test imports none of it. The `pinned (deploy floor)` leg passes in 2m14s.

Locally the same base hangs in a different place — `decode . encode = identity on (seed, actions)` in `replay-test` — with a let-go built from current `main`, allocating to 12.4 GB before we killed it. Clean `cc43ffb` with none of these commits reproduces that, and the same lg runs the suite to completion on rebased xsofy in 60 seconds, 444 tests and 3,066 assertions green.

So: two different tests, two different current let-go builds, both hanging on this base, and neither reachable from anything here. We have not chased the root cause, and the two hang sites do not share an obvious one — the replay test is pure encode and decode with no world construction. What we can say is that the base and a current let-go do not get along, and that rebasing clears it.

The last green run on this branch is 2026-07-05, so its passing checks predate whatever changed.

main_gl.lg
```diff
@@ -123,20 +123,44 @@
...
-                                               "left" "[D"
-                                               "right" "[C"
-                                               key))  ;; fallback to raw key
-    (string/starts-with? event-str "key-repeat:") (subs event-str 11)
+    (string/starts-with? event-str "key:escape") ""
+    (string/starts-with? event-str "key:")
+    (let [name (subs event-str 4)]
+      (or (arrow-seq name)
+          (when (> (count name) 1) name)))
+    (string/starts-with? event-str "key-repeat:")
```
nnunley, line 160: Perhaps a struct/enum-like thing would be better than the string matching. The pattern is established, however, so not a blocker. But worth considering in cleanup.

main_gl.lg
```diff
@@ -159,6 +183,58 @@
...
+   :int-sum 0.0 :last-t nil :bg 0 :glyphs 0 :total 0})
+
+(defn perf-note-verts!
+  "Record the batch sizes the current frame emitted, so a culling change shows
+   up as vertices, not only as milliseconds."
+  [perf* bg glyphs]
+  (when perf*
+    (swap! perf* assoc :bg bg :glyphs glyphs)))
+
+(defn perf-record!
```
nnunley, line 209: This seems like it might be nice to have in let-go as part of our tooling

nnunley (approved):

nooga (approved):
Reviewed via automated multi-angle analysis (line-by-line, removed-behavior, cross-file trace, cleanup, efficiency, altitude, CLAUDE.md conventions), with the diff correctly rebased against this PR's actual base (`gl-frontend`, not `main`).

**Verdict: approve.** The perf win is real and verified (2,370→40 culled cells at seed 42, ~400ms→~18ms cpu/frame, byte-identical rendered output). The two input-parsing fixes are also verified correct and consistent against both glplat backends (native GLFW and ebiten/WASM) — this closes a real "every keypress moves two tiles" bug.

Non-blocking notes:

1. `main_gl.lg` — the new `XSOFY_GL_PERF` instrumentation calls `perf-record!` every frame regardless of mode, but only `render-top-down`/`render-first-person` call `perf-note-verts!`. A sample window spanning a title→top-down or top-down→falling→fp transition will report real cpu/interval numbers next to stale/zeroed vertex counts. Dev-only debug tooling, low impact.
2. `parse-glplat-event`'s `key:` and `key-repeat:` branches duplicate the same `(or (arrow-seq name) ...)` logic verbatim — worth a one-line shared helper.
3. The two input-parsing fixes (arrow-key sequences, double-dispatch) ship with no automated test coverage despite the project's `test` namespace convention; the PR body already enumerates the exact cases manually verified, which would translate directly into a `deftest`.

Confirmed the failing "latest (upstream smoke test)" CI check is a pre-existing hang in `replay-test` reproducible on the unmodified `gl-frontend` base (already fixed on current `main` per the PR description and the raw CI log) — not caused by this diff.

nooga:
Checked the current CI run for this PR: `latest (upstream smoke test)` still fails, `pinned (deploy floor)` passes. Pulled the log — it's the same hang in `play-step-dispatches-explicit-action`/`replay-test` that both prior reviews (mine and nnunley's context) already traced to the `gl-frontend` base, reproducible on the unmodified base with no changes from this diff. `latest (upstream smoke test)` is the optional/upstream-tracking lane, not the required one, and `mergeable`/`mergeStateStatus` on this PR confirm there's no hard block — the UNSTABLE state is just that one non-required check.

Given both reviewers already approved and the failure is pre-existing on `gl-frontend`, this is safe to merge as-is. Fixing the base hang should happen separately, not as a condition of this PR.

PR #191, opened 2026-09-06
build(wasm): emit main.wasm externally and boot it with instantiateStreaming

mparrett:
## What changes

One flag in the shared build action: `lg -w` gains `-w-wasm external`, so the bundle ships a small `index.html` plus a separate `main.wasm` that the loader fetches and streams, instead of a single HTML file with a gzip-base64 payload baked into it.

`lg -w` has supported this since let-go#246. Nothing else in the chain moves.

## Why

Boot CPU. The inline path re-runs `atob`, then `DecompressionStream`, then instantiate on **every** load. `instantiateStreaming` compiles the module while it downloads, and the browser can reuse its compiled-module cache afterwards.

It is not a size win, and the numbers say so plainly. External ships **3.7% more** on the wire: 5,866,730 B in one request against 6,085,397 B in two, both gzipped by Pages. It still finishes in a seventh of the time on a warm load, because the cost was never transfer.

## Measurement

Two GitHub Pages deploys differing by exactly this commit, built from the same source. 14 rounds, first 4 discarded, arms alternating order each round so neither systematically eats the cold-CDN-edge penalty. Time from navigation to the xterm grid existing:

| | inline | external | ratio |
|---|---:|---:|---:|
| cold | 952 ms (905-1010) | 272 ms (228-345) | **0.286x** |
| warm | 331 ms (311-890) | 45 ms (37-58) | **0.136x** |

Per-round ranges are disjoint on both rows.

An A/A control on the same harness — identical URLs in both arms — put the warm noise floor at 2.8% and the cold floor at 17%. The warm effect is thirty times its floor; the cold effect is four times its own. An earlier version of the harness ran one arm first every round and reported a 44% difference between identical URLs, which is what prompted the alternation and the control.

## What was checked

- `inject_shell.lg` and `patch_wasm_coi.lg` both still find their markers under external mode and patch cleanly.
- Pages serves `main.wasm` as `content-type: application/wasm` with `content-encoding: gzip`, so streaming compile is actually available rather than silently falling back.
- The deployed bundle boots with `crossOriginIsolated === true` through the service-worker shim, with no console or page errors, and plays.
- `tests/browser/smoke.mjs` already maps `.wasm` to `application/wasm`, so the smoke gate needed no change.

## Scope

The cold-load advantage depends on the server compressing and setting the right content type. GitHub Pages does both, which is what production uses. A server that does neither can reverse the cold result; the warm result does not depend on it.

## Related CI fixes

The current upstream-latest failure is a stale `let-go` cache/toolchain issue, not a failure of the external-WASM change:

- Cache behavior originally introduced in #163
- Draft cache-key correction in #193
- Upstream cached-binary robustness fix in nooga/let-go#837

### DIFF
--- a/.github/actions/build-wasm/action.yml
+++ b/.github/actions/build-wasm/action.yml
@@ -24,11 +24,27 @@ runs:
   steps:
     # 1. Shell-less bundle: -w-shell none emits glue-only HTML exposing
     #    window.LetGoHost, so a project can inject its own client shell (#245).
+    #
+    #    -w-wasm external emits main.wasm alongside a small index.html and
+    #    boots it with instantiateStreaming, instead of baking a gzip-base64
+    #    payload into the HTML.
+    #
+    #    This is a boot-CPU win, not a bytes win. The inline path re-runs atob
+    #    + DecompressionStream + instantiate on every load; streaming compiles
+    #    during the download. Measured on two GitHub Pages deploys differing
+    #    only by this flag, 10 scored rounds each, time to the xterm grid:
+    #
+    #      cold  952ms -> 272ms  (0.286x)
+    #      warm  331ms ->  45ms  (0.136x)
+    #
+    #    External ships 3.7% MORE on the wire (5,866,730 B in one request
+    #    against 6,085,397 B in two, both gzipped by Pages) and still finishes
+    #    in a seventh of the time warm.
     - name: Build WASM (shell-less)
       shell: bash
       run: |
         if [ -n "${{ inputs.letgo-src }}" ]; then export LETGO_SRC="${{ inputs.letgo-src }}"; fi
-        let-go -w "${{ inputs.dist }}" -w-shell none main.lg
+        let-go -w "${{ inputs.dist }}" -w-shell none -w-wasm external main.lg
 
     # 2. Inject xsofy's own shell: rune font (Fairfax HD, OFL) + ?font=system.
     #    Without this the bundle ships let-go's generic shell (no rune face,
--- a/.github/actions/build-wasm/action.yml
+++ b/.github/actions/build-wasm/action.yml
@@ -24,22 +24,10 @@ runs:
   steps:
     # 1. Shell-less bundle: -w-shell none emits glue-only HTML exposing
     #    window.LetGoHost, so a project can inject its own client shell (#245).
-    #
-    #    -w-wasm external emits main.wasm alongside a small index.html and
-    #    boots it with instantiateStreaming, instead of baking a gzip-base64
-    #    payload into the HTML.
-    #
-    #    This is a boot-CPU win, not a bytes win. The inline path re-runs atob
-    #    + DecompressionStream + instantiate on every load; streaming compiles
-    #    during the download. Measured on two GitHub Pages deploys differing
-    #    only by this flag, 10 scored rounds each, time to the xterm grid:
-    #
-    #      cold  952ms -> 272ms  (0.286x)
-    #      warm  331ms ->  45ms  (0.136x)
-    #
-    #    External ships 3.7% MORE on the wire (5,866,730 B in one request
-    #    against 6,085,397 B in two, both gzipped by Pages) and still finishes
-    #    in a seventh of the time warm.
+    #    -w-wasm external emits main.wasm beside the HTML so the loader streams
+    #    and compiles it, rather than re-running atob + DecompressionStream +
+    #    instantiate on every load. A boot-CPU win, not a size win; external is
+    #    slightly larger on the wire. Timings in nooga/xsofy#191.
     - name: Build WASM (shell-less)
       shell: bash
       run: |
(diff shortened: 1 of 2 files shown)


.github/actions/build-wasm/action.yml
```diff
@@ -24,11 +24,27 @@ runs:
   steps:
     # 1. Shell-less bundle: -w-shell none emits glue-only HTML exposing
     #    window.LetGoHost, so a project can inject its own client shell (#245).
+    #
```
nnunley: Remove the devlog comment. THis captures metrics that will drift over time.

If this really needs to be captured and surfaced, add a docs/journal/ directory, and add dated notes there.

mparrett: Resolved in 9349d25 — the timings are out of the action comment. They stay in the commit message and the PR body, which are dated and cannot drift out of sync with the code. The comment keeps only the part that stays true: external delivery streams and compiles during the download instead of re-running atob + DecompressionStream + instantiate on every load, which is a boot-CPU win rather than a size win.

I did not add `docs/journal/`, since git already carries this measurement with a date attached. If you would rather have it in-tree as a doc, I will add it.

nooga (changes requested):
Reviewed via automated multi-angle analysis, cross-checked against the pinned/latest let-go binaries and actual CI logs.

The change itself is correct: `-w-wasm external` is a supported flag at the pinned let-go commit, there's a golden test for this exact `-w-shell none` + `-w-wasm external` combination, and the marker strings `tools/inject_shell.lg`/`tools/patch_wasm_coi.lg` depend on are present at that pin. Full suite passes (328 tests / 2857 assertions).

**Requesting changes on one gap before merge:**

- `tests/e2e/static-server.js:12` has no `.wasm → application/wasm` entry in its MIME map. `main.wasm` (new in this PR) is served as `application/octet-stream` in e2e CI, so `WebAssembly.instantiateStreaming` throws and the loader silently falls back to the buffer+`instantiate` path — meaning e2e CI never actually exercises the streaming boot path this PR exists to ship, including the perf win the PR description measures. (`tests/browser/smoke.mjs` already has the correct mapping, so this harness has drifted from that one.) One-line fix: add `'.wasm': 'application/wasm'` to the `TYPES` map.

Non-blocking note: `main.wasm` now ships as a separate, unhashed, constant-URL file (previously inlined). `deploy-pages.yml` publishes with `keep_files: true` and no cache-busting, so `index.html` and `main.wasm` could in principle skew independently across deploys if a future change makes them ABI-incompatible. Not a blocker, just worth being aware of.

Also: I pulled the actual CI log for the failing "latest (upstream smoke test)" check — it's `wasm_exec.js not found in GOROOT` on Go 1.26.7 against upstream/latest let-go, in the optional smoke lane, unrelated to this PR's diff. Doesn't block this PR (the required "pinned (deploy floor)" lane passes), just flagging since it doesn't obviously match the cache-staleness story in the PR description (cross-referenced from #193).

mparrett:
Addressed the requested E2E MIME gap in 70214ee: `tests/e2e/static-server.js` now maps `.wasm` to `application/wasm`, so the worker-mode E2E server exercises streaming compilation instead of the `arrayBuffer` fallback.

Targeted verification:
- `node --check tests/e2e/static-server.js`
- `git diff --check`
- live header probe: HTTP 200, `Content-Type: application/wasm`, with COOP/COEP unchanged

mparrett:
Both asks are addressed as of 70214ee (Sep 11): `tests/e2e/static-server.js` maps `.wasm` to `application/wasm`, and the action comment keeps the rationale with the numbers in the commit message and the PR body.

Your review validated `-w-wasm external` and the `inject_shell.lg` / `patch_wasm_coi.lg` marker strings at the pinned commit. #195 moves that pin to v1.13.0, so I re-ran the check at the new version rather than assume it carries.

Built this branch's tree with lg v1.13.0 (369e2a69), all four steps of the composite action:

- `-w-shell none -w-wasm external` emits `main.wasm` beside `index.html`, and the loader calls `instantiateStreaming`
- `inject_shell.lg` finds its markers; the injected shell has the `#app` mount, the Fairfax face, and `?font=system`
- `patch_wasm_coi.lg` patches cleanly
- `write-build-info.sh` writes `build-info.json`

Headless boot of that bundle, served with COOP/COEP and `application/wasm`: `crossOriginIsolated` true, title screen reached, keypress lands on the game screen, no console errors.

The two PRs are order-independent, so whichever lands second needs no change.

On the unhashed `main.wasm`: `index.html` and `main.wasm` are written by the same build step and published together, so they can skew only if a deploy half-fails. If that ever bites, the fix is a content hash in the filename rather than anything in this PR.

nooga (approved):
Re-review. The requested gap is closed: `tests/e2e/static-server.js:12` now maps `.wasm` → `application/wasm` (added in 70214ee), matching `tests/browser/smoke.mjs`'s existing mapping. e2e CI now actually exercises `WebAssembly.instantiateStreaming` instead of silently falling back to the buffer+`instantiate` path, so the perf win this PR ships is the thing under test.

Re-checked the rest holds from the first pass: `-w-wasm external` is valid at the pinned let-go commit, the golden test for `-w-shell none` + `-w-wasm external` still covers this combination, and the `deploy-pages.yml` cache-busting note from before remains a non-blocking follow-up, not a blocker.

Approving.

PR #192, opened 2026-09-06
feat(hash): back the seed hash with murmur3 so the game builds for TinyGo

mparrett:
## The bug

xsofy's determinism layer called let-go's `xxh3` namespace. That namespace is gated `!tinygo` upstream — xxh3's bindings are reflect-boxed and its implementation is assembly-only on arm64/amd64 with no purego mode — so under TinyGo the call returns nil and the game dies at its first dice roll, on the title screen:

```
FATAL ERROR: TypeError: nil is not a function
  new-game <- show-title-screen <- generate-title <- pick <- int-in <- xxh3-64
```

## The change

let-go's `murmur3` namespace is registered on every build and mirrors the same `Hash` / `HashSeed` / `HashString` / `HashStringSeed` surface, so the swap is one call site. With it the game builds and plays under TinyGo wasm at 2.0 MB, against 8.4 MB for the stock-Go build.

`xxh3-64` is renamed to `seed-hash`, including the fully-qualified calls the salt-folding macros emit. A function named for an algorithm it no longer runs would be worse than the original bug.

## What this buys

murmur3 masks its results to 31 bits, so a value is identical on 64-bit native and 32-bit wasm. **One seed now produces one world on every target.**

That is stronger than what xsofy had. TinyGo's wasm `int` is 32-bit, so a 64-bit hash made the wasm build its own determinism domain — the reason the TinyGo lane was previously considered a demo rather than a target. Verified by diffing the rendered dungeon at a fixed seed between the TinyGo wasm build and the stock-Go wasm build: byte-identical, same sha256.

## What it costs

Seeds from v0.0.2 and earlier no longer name the worlds they used to.

This is accepted knowingly and recorded in `CHANGELOG.md`. Determinism itself is intact — a seed is still fully reproducible, and now reproducible across targets, which it was not before. What changed is which world a given seed names. Re-pinning published seeds is deferred rather than solved.

## What 31 bits changes

The narrower domain has two consequences the code handles explicitly.

**The commit DAG fails loudly on a reused id.** `dag/commit` addresses nodes by the world's rolling seed, so a 31-bit space makes a collision plausible where 64 bits made it negligible. Committing over an existing node whose `(parent, action, turn)` differs now throws rather than overwriting it and corrupting the replay chain that depended on it. Two things can land on one id: the hash width, and `dispatch` folding only the action *type* into the seed — so two parameterized actions of one type, from the same parent at the same turn, collide with no hash collision at all. No current call site reaches either, since forward play commits from the live HEAD and the unwind branch carries the current chain's seed; `branch` or `undo` wired to a player action would.

**External seeds are constrained to the portable domain.** `?seed=` and `XSOFY_SEED` are taken modulo 2^31 at the boundary, so the seed the game prints is always one that replays the same run on every target. In-range seeds are untouched, only a negative or `>= 2^31` seed is remapped, and `*boot-seed*` is a test hook that bypasses it.

## Tests

The xxh3 golden matrix served two purposes: a drift regression guard, and a canonical-algorithm proof cross-checked against `xxh3sum`. The first still matters, and its 84 goldens are regenerated against the new hash. The second no longer applies — reference verification for murmur3 lives upstream in let-go's `hash_murmur3_test.go`, against Appleby's reference implementation and twmb/murmur3.

Added `seed-hash-is-width-portable`, which hashes every fixture live and asserts the result is non-negative and below 2^31. Nothing previously tested the property that the whole cross-target determinism claim rests on.

The collision guard and the seed normalization each get targeted coverage, including a case pinning that re-committing identical metadata is not a collision. The seeded browser regression carries the same golden title as `xsofy/test/title_test.lg`; both move with the hash.

Suite: 332 tests, 2947 assertions, 0 fail (2857 assertions before).

### DIFF
--- a/.let-go-version
+++ b/.let-go-version
@@ -1,14 +1,23 @@
 # let-go version for building xsofy — single source of truth.
 # deploy-pages.yml and the test.yml floor lane both resolve the version from here.
 #
-# Pinned to a commit (not a v-tag) because the shell now binds the #app host
-# mount from nooga/let-go#378, which no release carries yet:
+# Pinned to a commit (not a v-tag) because the pieces xsofy needs keep landing
+# ahead of let-go's release cadence:
 #   - #378 renames the -w-shell none host mount #terminal -> #app and ships it
 #     visible; the shell (tools/xsofy-shell.html) binds #app.
+#   - #522 adds the `murmur3` namespace, which xsofy.hash/seed-hash calls. It
+#     merged 2026-07-23, three days after v1.12.2 (2026-07-20), so no release
+#     tag carries it. This is the constraint that moved the pin.
 # Still includes #245 (client-owned shell) and #339 (js/url-param host seam)
 # from v1.11.1.
 #
+# This commit is the one xsofy was verified against on 2026-09-05: full test
+# suite green (328 tests, 2942 assertions), stock-Go and TinyGo wasm bundles
+# both built via the shipping three-step (-w-shell none -> inject_shell.lg ->
+# patch_wasm_coi.lg), both boot and play, and the two produce byte-identical
+# worlds at a fixed seed.
+#
 # NOTE: this is a SHA, not a release tag, so the build must check out + build
 # let-go at this commit — a stale prebuilt `lg` silently emits the old #terminal
-# frame. Re-pin to a release tag once let-go ships #378.
-cf195a67fd8aa43286319519c2512363caa96a52
+# frame. Re-pin to a release tag once let-go ships a release containing #522.
+091111861106b48b008fcf7ac3c4ed451410821b
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,16 @@ player sees from the previous release; intra-cycle churn (introduced and fixed
 between tags) is omitted. Web-only items are marked `(web)`; everything else
 applies to both builds.
 
+## Unreleased
+
+- **Seeds from v0.0.2 and earlier no longer reproduce their old worlds.** The
+  deterministic seed hash moved from xxh3 to murmur3 so the game can build for
+  TinyGo, which has no xxh3 namespace at all. A given seed is still fully
+  deterministic and now produces the *same* world on 64-bit native and 32-bit
+  wasm alike, which it did not before; but the world a pre-existing seed names
+  has changed. Accepted knowingly: seed sharing predates any released run
+  worth preserving, and re-pinning seeds is deferred rather than solved.
+
 ## v0.0.2
 
 - Runes render in the browser terminal — v0.0.1 showed tofu/overflow, because the
--- a/tests/e2e/regression.spec.js
+++ b/tests/e2e/regression.spec.js
@@ -41,9 +41,10 @@ function watchConsole(page) {
 
 // Golden title for seed 12345 (computed natively from the word lists). Seeing
 // it render in-browser proves the whole determinism chain: ?seed= reached the
-// game via go.env, AND wasm-xxh3 == native-xxh3 (a divergence in either would
-// produce a different title).
-const TITLE = 'Chasms of Infinity';
+// game via go.env, AND the wasm seed hash == the native one (a divergence in
+// either would produce a different title). Must match the golden in
+// xsofy/test/title_test.lg — both move together when the seed hash changes.
+const TITLE = 'Cells of Nil';
 
 function waitForTermText(page, needle, timeout = 30000) {
   return page.waitForFunction(
@@ -51,7 +52,7 @@ function waitForTermText(page, needle, timeout = 30000) {
     needle, { timeout, polling: 200 });
 }
 
-test('seed 12345 reproduces the title (proves ?seed bridge + wasm xxh3 parity)', async ({ page }) => {
+test('seed 12345 reproduces the title (proves ?seed bridge + wasm hash parity)', async ({ page }) => {
   const srv = serve();
   const errors = watchConsole(page);
   try {
@@ -68,7 +69,7 @@ test('seed 12345 reproduces the title (proves ?seed bridge + wasm xxh3 parity)',
 // Regenerate with: lg -e '(require (quote [xsofy.replay :as r])) (println (r/encode 12345 [...]))'
 const REPLAY_CODE = 'AQAAAAAAADA5AgR3YWl0BGRvd24AAAADAAIBAQAB';
 
-test('?replay= plays back the run (decode + animated playback + xxh3 parity)', async ({ page }) => {
+test('?replay= plays back the run (decode + animated playback + hash parity)', async ({ page }) => {
   const srv = serve();
   const errors = watchConsole(page);
   try {
--- a/tests/e2e/regression.spec.js
+++ b/tests/e2e/regression.spec.js
@@ -67,7 +67,7 @@ test('seed 12345 reproduces the title (proves ?seed bridge + wasm hash parity)',
 
 // Replay code for seed 12345 + [:wait :wait :down :wait] (xsofy.replay/encode).
 // Regenerate with: lg -e '(require (quote [xsofy.replay :as r])) (println (r/encode 12345 [...]))'
-const REPLAY_CODE = 'AQAAAAAAADA5AgR3YWl0BGRvd24AAAADAAIBAQAB';
+const REPLAY_CODE = 'AgAAAAAAADA5AgR3YWl0BGRvd24AAAADAAIBAQAB';
 
 test('?replay= plays back the run (decode + animated playback + hash parity)', async ({ page }) => {
   const srv = serve();
--- a/xsofy/dag.lg
+++ b/xsofy/dag.lg
@@ -5,7 +5,7 @@
    the keypress that produced it; any past world is regenerated on demand by
    replaying its action chain from the origin checkpoint (~0.35s for a level —
    cheap because the world is pure data over (seed-input, action-log)). This
-   avoids storing or content-hashing a full world every turn (~65ms each).
+   avoids storing or content-hashing a full world every turn.
 
    - commit-id  = the world's rolling :seed hash (already advanced per turn by
                   det/advance; O(1), no per-turn full-world walk).
@@ -21,8 +21,11 @@
             [xsofy.det :as det]))
 
 (def default-budget
-  "Retention budget: max commits kept on HEAD's chain before GC advances the
-   origin checkpoint and forgets older history."
+  "Retention budget: DECLARED SHAPE ONLY — not yet enforced. Today the only
+   pruning is reachability GC (see `gc`), which drops abandoned branches but
+   never trims a HEAD chain that is still growing. Budget-based retention
+   (evicting the oldest reachable commits once :max-nodes is exceeded) is
+   deferred; see `gc`'s docstring for what that would require."
   {:max-nodes 1000 :evict-policy :oldest-by-turn})
 
 (defn commit-id
@@ -33,11 +36,12 @@
   (or (:seed world) 0))
 
 (defn checkpoint-hash
-  "OPTIONAL full content hash: seed-hash of the world's canonical bytes (~65ms on
-   a full floor — see tools/). NOT for per-turn identity (that's commit-id).
-   Compute only when non-history-derived state is merged into a world (save/load
-   reconstruction, external injection) to re-anchor/verify content :seed can't
-   account for."
+  "OPTIONAL full content hash: seed-hash of the world's canonical bytes. NOT for
+   per-turn identity (that's commit-id) — walking and encoding every entity is
+   far more expensive than the O(1) rolling :seed hash, so this is not meant to
+   run every turn. Compute only when non-history-derived state is merged into a
+   world (save/load reconstruction, external injection) to re-anchor/verify
+   content :seed can't account for."
   [world]
   (hash/seed-hash (ser/canonical-bytes world)))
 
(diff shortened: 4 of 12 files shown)


mparrett:
Marking this a draft: it depends on a let-go change that has never shipped in a release.

`pkg/rt/hash_murmur3.go` landed in nooga/let-go#522 on 2026-07-23. The latest let-go release, v1.12.2, is dated 2026-07-20 — three days earlier — so no tag contains it. xsofy's CI builds from `.let-go-version`, currently pinned to `cf195a67` (2026-07-06), and fails as it should:

```
CompileError: Can't resolve murmur3/HashSeed in this context
  --> xsofy/test/console_test.lg:1:1
```

That is the pin doing its job. I had verified the change against a let-go built from `main`, which has murmur3, and missed that the pinned build does not.

Three ways forward, none of which should be picked quietly:

1. **A let-go release containing #522.** Cleanest: it also lets `.let-go-version` move from a SHA to a tag, which the file's own comment asks for once #378 ships. #378 is already in v1.12.2.
2. **Bump the pin to a SHA with murmur3.** The minimum is `826b5f97` (#522's merge), 128 commits past the current pin; current `main` is 210. Consistent with existing practice — xsofy already pins a SHA — but it pulls two months of runtime changes in alongside a hash swap, and I would rather those land as separate reviewable steps.
3. **Wait.** The TinyGo lane is not on any critical path.

Not proposing a conditional fallback to xxh3 when murmur3 is absent: that would make the world a seed names depend on which let-go built the bundle, which is worse than the bug this fixes.

The change itself stands on its own — full suite green locally (328 tests, 2942 pass, 0 fail), and the TinyGo and stock-Go wasm builds produce byte-identical worlds at a fixed seed. It is only the pin that blocks it.

mparrett:
Unblocked by moving the pin, which was already a commit pin rather than a tag — `.let-go-version` has pointed at a SHA since #378, and the file explains why. Moving it forward is the ordinary maintenance action here, not a departure. My earlier comment framed this as a larger decision than it is.

`cf195a67` (2026-07-06) → `091111861106` (2026-09-04).

**`pinned (deploy floor)` is green.** That is the `required: true` lane and the one the Pages deploy resolves from.

**`latest (upstream smoke test)` still fails, with the same `Can't resolve murmur3/HashSeed`.** That is expected and correct rather than something left broken: `@latest` resolves to the newest release *tag*, and no release contains #522. The lane is `required: false` with `continue-on-error`, and a red result there is exactly the signal it exists to give — this does not build against a released let-go yet. It goes green on its own the first time a release carries #522, with no change here.

I picked `091111861106` over the minimum that would satisfy the dependency (#522's own merge, `826b5f97`) because it is the commit the change was actually verified against: full suite at 328 tests / 2942 assertions, stock-Go and TinyGo wasm bundles both built through the shipping three-step and both booting and playing, and the two producing byte-identical worlds at a fixed seed. The smaller jump would have been less change but entirely unverified.

Ready for review.

nooga (changes requested):
Reviewed via automated multi-angle analysis (8 finder angles + verification pass, run twice independently), a full pinned-toolchain test run (328 tests / 2942 assertions, 0 failures), and manual verification of the murmur3 backend against the pinned let-go commit.

The core motivation and mechanism are sound — xxh3's let-go bindings are reflect-boxed/assembly-only and unavailable under TinyGo, so murmur3 (registered on every target, masked to 31 bits for native/wasm parity) is the right fix for the stated problem.

**Requesting changes on one point before merge** — multiple independent review passes converged on this:

- `xsofy/dag.lg:28` — `commit-id`/`checkpoint-hash` use the seed-hash directly as an *unguarded* map key in `:nodes` (`assoc-in [:nodes id] {...}`, no existence check). The hash width backing this address space just dropped from 64-bit (xxh3) to 31-bit (murmur3, by design for wasm parity). Verified: `:max-nodes`/`:evict-policy` are declared in `default-budget` but **never enforced anywhere** in the codebase — no eviction call site exists — so `:nodes` grows unbounded for a session's lifetime, and birthday-bound collision probability grows as O(n²) against a 2^31 space (~46k nodes ⇒ non-trivial odds). A collision would silently overwrite an existing node's `{:parent :action :turn}`, corrupting undo/replay with no error raised. Please either (a) add a guard against overwriting an existing `:nodes` entry on hash collision, (b) widen the DAG's key beyond the raw seed hash, or (c) if unbounded growth within a session is already an accepted constraint, say so explicitly in the docstring so it's a documented tradeoff rather than a silent gap.

Related, lower-priority items worth folding into the same pass:

- `xsofy/seed.lg:24` — `boot-seed` accepts an unclamped player-supplied `?seed=`/`XSOFY_SEED` integer. Above 2^31-1, this could hit `BigInt` vs `Int` representation differences between native and TinyGo-wasm in the seed-hash's bit-shift path — exactly the cross-platform-determinism scenario this PR is meant to guarantee, but outside what the new tests cover.
- Doc drift, easy fixes: `xsofy/hash.lg:2,92` still says "64-bit" (actual output is 31-bit); `xsofy/det.lg:79`'s `int-in` bias claim (`span / 2^64`) is now off by 2^33; stale "xxh3-64" references remain in `xsofy/test/hash_test.lg`'s ns docstring and a couple of comments.
- Test quality: `seed-hash-is-width-portable` (`xsofy/test/hash_test.lg`) only asserts against the hardcoded golden literals — it never calls `h/seed-hash`, so it can't actually catch a width regression in the live implementation.
- `xsofy/dag.lg:36`'s `checkpoint-hash` docstring still claims "~65ms on a full floor" (carried forward, not re-verified). Benchmarked it at ~10.7s on a 79×30 world, almost entirely in `ser/canonical-bytes` (not touched by this PR, and `checkpoint-hash` has no production callers yet, so not a regression — but worth fixing before someone wires it into a synchronous save/load path on the strength of the stale number).

Separately observed during review, pre-existing on `main` (not introduced by this PR, flagging since it surfaced repeatedly across independent passes): `xsofy/console.lg`'s `(bench)` command still labels its output "ms" though `perf.lg` now returns microseconds, and `tools/xsofy-shell.html`'s `no-touch` UI state has no in-app recovery path for a non-dev touch-only session that lands in `off`. Worth separate follow-up issues, not part of this review.

mparrett:
Pushed 15b554d, taking the five points in order.

**1. DAG collision guard.** `dag/commit` now compares an existing node at the same id against the (parent, action, turn) it is about to write, and throws on a mismatch instead of overwriting. Added a targeted test, plus one pinning that re-committing identical metadata is not a collision.

Two things can reuse an id, not just the 31-bit hash: `dispatch` folds only the action *type* into the seed, reproducing an action's params from `:action-log` instead, so two parameterized actions of the same type from the same parent at the same turn address the same node with no hash collision involved. I checked all six `commit` call sites and neither path is reachable today: forward play commits from the live HEAD, and the unwind branch carries the current chain's seed. `dag/branch` or `dag/undo` wired to a player action would hit the second one deterministically. Widening the key is still the larger change you ruled out, so the guard only makes the failure loud.

**2. DAG documentation.** `default-budget` now states it is a declared shape, with budget retention deferred and reachability GC the only pruning that exists. Removed the `~65ms` `checkpoint-hash` figure.

**3. Hash documentation.** Reworded the "64-bit" and xxh3 leftovers, and moved the modulo-bias denominator in `det/int-in` from 2^64 to 2^31. I also removed the "xxh3-64 published vectors" comment block, which describes a reference-vector method the suite no longer uses, and the `str->bytes` helper it fed. That helper was already unused on `main` rather than orphaned by this PR, so say the word if you would rather I restore it and keep the diff to this PR's own leftovers.

**4. Width test.** `seed-hash-is-width-portable` computes the hash live instead of asserting over the pinned golden literals, so it can now fail on an implementation change.

**5. Seed normalization.** `?seed=` and `XSOFY_SEED` are taken modulo 2^31, so the seed the game prints always replays the same run on 64-bit native and 32-bit wasm alike. In-range seeds are untouched; only a negative or `>= 2^31` seed is remapped. `*boot-seed*` is a test hook and bypasses it.

The 31-bit domain also narrows `combine-bit-spread`, which draws 100 hashes and requires all of them distinct across 100 defprop sets. The birthday bound is about 2.3e-6 per set, so the property should fail spuriously roughly once in 4000 suite runs, a margin that was invisible at 64 bits. I left it documented at the test rather than changed, since the assertion is yours to loosen; accepting 99 distinct would remove the flake if you would rather not carry it.

Suite green as of 2026-09-10: 332 tests, 2947 assertions, 0 failures.

mparrett:
cc: @nnunley Touches your domain, any concerns?

mparrett:
Follow-up in 092c794: the collision test now uses `try`/`catch` instead of `(is (thrown? ...))`. `thrown?` postdates v1.12.2, so the new test carried its own incompatibility with the "latest" lane — masked behind the `murmur3/HashSeed` error today, and it would have surfaced as a fresh red the moment let-go cut a release carrying murmur3. `try`/`catch` works on both versions and is what the rest of the suite already uses.

nnunley:
I think breaking old replays is less desirable. Perhaps have a field for the hashing methods, with a missing field translating to the old hashing methods?

mparrett:
Correction to my earlier comment. I wrote that no current call site can reach the collision guard. That holds for one of the two routes and is wrong for the other.

The structural route is as I described: two parameterized actions of one type from the same parent, colliding because `dispatch` folds only the action type into the seed. Not reachable today.

The hash route is reachable in ordinary forward play. `:nodes` keeps every commit in a run — nothing prunes a linear chain, and `default-budget` is not enforced — so each turn writes a new 31-bit id into a growing set, which is a birthday problem against every earlier turn. `1 - exp(-T^2/2^32)` puts that near 2% at 10k turns and 9% at 20k. I measured the retention rather than assuming it: 200 turns of `:wait` leaves 201 nodes.

So the guard is not theoretical, and with it a collision becomes a crash mid-run where it used to be silent metadata corruption. Which of those is preferable depends on xsofy's typical run length, which you would judge better than I can.

Two things that keep this in proportion. It is a narrowing rather than a new design: `main` already keys the DAG on the seed, at 64 bits where the margin is `T^2/2^65`. And the durable fix is the identity rather than the hash — 31 bits is forced by let-go's `boxHash`, but keying nodes on a `[turn seed]` pair or a counter removes the class entirely without touching determinism. That is the wider-key change you ruled out for this PR, so I have left it alone and only corrected the docstring's claim (9aba291).

mparrett:
@nnunley What follows is Claude's, posted under my account — it made the `VERSION` change and checked the xxh3/TinyGo constraint against the code. The compatibility call at the end is a proposal, not a decision.

You're right that it breaks them, and the break is wider than replays — every `?seed=` names a different world too. The swap changes both the algorithm and the output width (xxh3-64 to murmur3 masked at 31 bits), so nothing derived from a seed survives it. `CHANGELOG.md` records that as a knowing cost rather than an oversight, but it is a real one.

I took the half of your suggestion that works today. Replay codes already carry a version byte, and the wire format reserved this case in as many words: "future versions reject v1 codes by default". `VERSION` is now 2, so an old code hits the existing "unknown code version" error instead of decoding cleanly and replaying a world it never recorded. That last behaviour was the worst option available, because it looks like it worked.

What I did not add is the fallback half, for three reasons.

It cannot run where it is needed. `interop_xxh3.go` is gated `!tinygo`, so a TinyGo build has no `xxh3` namespace at all. A "missing field means old hashing" path would be unimplementable on the one target this change exists to serve, and preserving it on stock-Go alone would mean the same code replays a world in the browser and errors in the small build.

`?seed=` has nowhere to put the field. It is a bare integer in a URL. Replay codes can carry a version; seeds cannot, short of changing the grammar. Defaulting a missing field to the old hash would make xxh3 permanent for every shared seed and murmur3 the opt-in, which inverts the change.

Two live hashes is a standing determinism cost. Both would have to stay bit-exact on every target indefinitely, and the matrix proving it doubles.

For completeness, one route does preserve every existing seed: making let-go's `Int` an `int64` would let xxh3-64 round-trip on 32-bit wasm unchanged. That was considered and set aside as too large a VM change rather than overlooked. If the compatibility is worth that price, it is a let-go conversation rather than an xsofy one.

My read is that the break is acceptable here, pre-1.0 and with seed sharing not yet having outlived a release. You have a longer view than I do of who is actually holding links, so I am happy to revisit it on that basis.

mparrett:
What follows is Claude's, posted under my account; the decision to close is mine.

Closing this in favour of fixing the width at the source. @nnunley was right about old replays, and the route that keeps them turned out to be a bounded let-go change once measured.

**What the swap was working around.** xsofy's TinyGo build had two problems, and murmur3 solved both by narrowing: let-go had no `xxh3` namespace under TinyGo, and let-go's `Int` followed the platform, so a 64-bit digest could not survive the trip into a 32-bit `Int`. That second one is a let-go bug, not a property of the hash. It also hid a `MaybeDowngrade` truncation on 32-bit hosts.

**What replaces it.** Two let-go changes:

- nooga/let-go#859 makes `Int` int64 on every host: the numeric core, opcode fast paths, reader literals, and the interop edges that narrowed a 64-bit producer through `int`. No change on 64-bit targets.
- nooga/let-go#860, stacked on it, registers the `xxh3` namespace on TinyGo wasm builds through the same non-reflect path murmur3 uses.

With both, xsofy `main` runs its xxh3-64 golden matrix and determinism tests under TinyGo wasi with no source change, 68 tests and 597 assertions passing bit for bit against native. Every existing seed and replay code keeps naming the world it names today, on every target.

**What this PR carried that is no longer needed.** The DAG collision guard existed because 31-bit ids made reuse plausible in ordinary play; at 64 bits the margin is back to T^2/2^65. The `?seed=` and `XSOFY_SEED` normalisation existed to keep external seeds inside the portable domain; the whole int64 range is portable now. The replay `VERSION` bump existed to reject codes that would decode into a different world; they no longer would. The CHANGELOG entry recorded a break that will not happen.

**What xsofy still needs, after the let-go side lands.** A `.let-go-version` bump to a commit carrying both changes, and nothing else. The pin bump this branch made (`cf195a67` to `091111861106`) is still the right kind of move and can go in with that.

**Performance note, for the record.** The hash width does not move the TinyGo lane: 20k `det/int-in` rolls take the same time with 64-bit xxh3 seeds as with 31-bit murmur3 seeds. What does move it is TinyGo's collector: the same loop runs 20x faster with collection disabled. That is a separate piece of work.

The branch is archived on the fork before deletion.

PR #193, opened 2026-09-09
ci: key lg cache on exact Go version

mparrett:
## Summary

- resolve the exact active Go version alongside GOPATH
- include that version in the cached `let-go` binary key
- keep test, deploy, and restore-only preview workflows on the same cache identity

## Why

#163 introduced the `let-go` binary cache and correctly derived GOPATH from `go env` so the cache path follows the install target. The cache key still identified the toolchain only as `go1.26`, however, so every Go 1.26 patch release shared one compiled-binary cache.

#191 exposed the consequence: a runner using Go 1.26.8 restored a `let-go` binary built under Go 1.26.7. The binary retained that vanished build-time GOROOT and its WASM smoke test could not find `wasm_exec.js`.

This change derives `LG_CACHE_GO_VERSION` from `go env GOVERSION` and includes it in all three keys. A Go patch update now misses once, rebuilds the binary with the active toolchain, and seeds matching deploy/preview caches.

The companion upstream robustness fix is nooga/let-go#837. That fix makes cached or relocated `let-go` binaries follow the Go toolchain actually selected for the WASM build; this PR independently makes the cache identity describe the binary's full build toolchain.

## Related

- Cache introduction and GOPATH discussion: #163
- Failure that exposed the missing key input: #191
- Upstream relocatability fix: nooga/let-go#837

## Verification

- `actionlint .github/workflows/test.yml .github/workflows/deploy-pages.yml .github/workflows/pr-preview.yml`
- `git diff --check`

### DIFF
--- a/.github/workflows/deploy-pages.yml
+++ b/.github/workflows/deploy-pages.yml
@@ -49,15 +49,18 @@ jobs:
       # to be re-verified against a run to know it is right.
       - name: Resolve cache paths
         run: |
-          echo "GOPATH=$(go env GOPATH)" >> "$GITHUB_ENV"
-          echo "PLAYWRIGHT_BROWSERS_PATH=$HOME/.cache/ms-playwright" >> "$GITHUB_ENV"
+          {
+            echo "GOPATH=$(go env GOPATH)"
+            echo "LG_CACHE_GO_VERSION=$(go env GOVERSION)"
+            echo "PLAYWRIGHT_BROWSERS_PATH=$HOME/.cache/ms-playwright"
+          } >> "$GITHUB_ENV"
 
       - name: Cache lg binary
         id: lgcache
         uses: actions/cache@v4
         with:
           path: ${{ env.GOPATH }}/bin/let-go
-          key: lg-bin-${{ runner.os }}-go1.26-${{ steps.letgo.outputs.version }}
+          key: lg-bin-${{ runner.os }}-${{ env.LG_CACHE_GO_VERSION }}-${{ steps.letgo.outputs.version }}
 
       - name: Install let-go
         if: steps.lgcache.outputs.cache-hit != 'true'
--- a/.github/workflows/pr-preview.yml
+++ b/.github/workflows/pr-preview.yml
@@ -69,7 +69,9 @@ jobs:
           } >> "$GITHUB_ENV"
 
       # pull_request_target runs may restore default-branch caches but cannot
-      # save a miss. Trusted deploy/manual runs seed this cache.
+      # save a miss. Trusted deploy/manual runs seed this cache. After a Go
+      # patch bump the new key is empty until test.yml on main or a deploy run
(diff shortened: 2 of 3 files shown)


nooga (approved):
Reviewed via automated multi-angle analysis plus verification against the pinned and latest let-go toolchains.

**Verdict: approve.** The cache-key fix is mechanically correct and consistently applied across all three workflows — no leftover hardcoded `go1.26` reference remains anywhere.

Two non-blocking comment-accuracy nits:

1. `pr-preview.yml`'s cache lane is restore-only (`actions/cache/restore@v4`, no write-back) — after a Go patch bump, this lane can't reseed the new key itself. In practice `test.yml`'s read-write cache reseeds it on the very next PR, so this is self-healing, but worth a one-line comment noting the dependency.
2. The `GOVERSION` comment implies it captures the *exact* toolchain that built the binary. In principle `GOTOOLCHAIN=auto` could switch toolchains mid-build if a future let-go `go.mod` declares a newer `toolchain` directive than the runner's Go — pre-install `go env GOVERSION` wouldn't see that. Checked: neither the pinned commit nor the latest let-go release declares a `toolchain` directive today, so this is inert, but worth knowing if it's ever added upstream.

Also verified: the actual CI failure on the "latest (upstream smoke test)" lane referenced from #191 has a different root cause than the cache-staleness scenario this PR's description attributes to it (see my review comment on #191) — doesn't affect this PR's correctness, just flagging so the two PR descriptions aren't cross-cited incorrectly.

mparrett:
Reopened. This closed automatically at 14:03 UTC when nooga/let-go#837 merged: its body links this PR as "Companion cache-identity fix: nooga/xsofy#193", and GitHub read the "fix:" before the reference as a closing keyword. The change here is still needed on its own.

mparrett:
Both nits are in 1d3444e.

1. `pr-preview.yml` now says the restore-only lane rebuilds from source after a Go patch bump until `test.yml` on main or a deploy run writes the new key.
2. The `GOVERSION` comment now says it is the toolchain setup-go activated, and that under `GOTOOLCHAIN=auto` `go install` switches to a newer `toolchain` line in let-go's `go.mod`. One correction to the check behind the nit: v1.12.2 has no `toolchain` line, but let-go `main` declares `toolchain go1.26.5`. That is below the runner's 1.26.8, so the key is still accurate today; the comment carries the version and date so it can be re-checked.

On the #191 cross-reference: the description's story matches the log. In run 34273241540, the latest lane set up Go with `GOROOT='/opt/hostedtoolcache/go/1.26.8/x64'`, then restored `lg-bin-Linux-go1.26-v1.12.2` from cache, then failed with `wasm_exec.js not found in GOROOT (/opt/hostedtoolcache/go/1.26.7/x64)`. The 1.26.7 in the error is the restored binary's build-time GOROOT, not the runner's toolchain, which is the staleness this key change prevents.

PR #194, opened 2026-09-12
perf(hash): encode salt terms into an accumulator

mparrett:
The det hot path allocated 96 objects per roll, and 94 of them were salt encoding: `(assoc world :seed ...)` accounts for the other 2. Every layer built a vector that the next layer copied through `into` and discarded.

`int->8-be` was the worst, at 18 allocations to produce 8 bytes. Written as `(mapv f (range 8))` it allocates a range seq and a closure before it allocates the result, and neither the IR optimizer nor the Go lowerer removes them. Lowered, the closure boxes to a native fn, the call goes through a `clojure.core/mapv` trampoline, and since the closure's parameter is untyped, every operation inside it runs through the boxed `rt.*Value` helpers. The `loop`/`recur` form lowers to native `int64` locals and arithmetic instead. `utf8-bytes` had the same shape, reducing a closure over the string and building a 1-4 element vector per codepoint.

## What changed

The encoders take an accumulator and append to it:

- `conj-int->8-be` replaces the intermediate 8-element vector.
- `encode-salt-into` replaces the per-term vector, and the vector arm folds its elements into the same accumulator.
- `conj-codepoint->utf8` appends a codepoint's bytes directly, and `codepoint->utf8-len` reports the byte count without building them, which is what lets the length byte be written before the bytes rather than after materializing them.
- `with-folded-salt` seeds one accumulator with the header and the folded prefix, rather than building a tail vector and copying it through two `into` calls.

`encode-salt`, `int->8-be`, and `codepoint->utf8` remain as named functions, each delegating to its accumulator version, because the JVM cross-validation twin in `test/hash_test_jvm.clj` mirrors them directly. That file is unchanged, so it stays an independent reference. The two `reduce` helpers are named rather than inline closures for the same reason the encoders changed shape.

## Measurements

Allocations per roll, 96 to 55. `allocs/op` is exact, and was identical across three runs:

```
BenchmarkFoldedRollBefore    96 allocs/op
BenchmarkFoldedRollAfter     55 allocs/op
```

Encoding a six-character keyword name, 72 allocations to 30.

Turn time, as medians of interleaved runs alternating which side goes first, each with an A/A control running the same build in both arms:

| lane | before | after | | A/A floor |
|---|---|---|---|---|
| native, 20k `det/int-in` rolls | 0.325s | 0.210s | 35% faster | 1.2% |
| TinyGo wasi under `wasmtime`, same | 30.9s | 19.1s | 38% faster | 0.2% |
| browser, engine bench over 3 fixtures | 64.8s | 51.7s | 20% faster | 1.9% |

The browser figure is per-turn engine cost inside the wasm runtime, driving `xsofy.perf`'s shared core with no render, so it is comparable to `bench/native.lg`. Average turn time across the three fixtures goes from about 64ms to about 51ms; native is about 9ms, so the wasm penalty stays the dominant term. The lanes with a conservative collector gain most, because allocation rate is what dominates there.

## Verification

The encoding is unchanged, which is the point: a shifted seed invalidates every published replay. Checked three ways.

The suite passes 330 tests and 2864 assertions, including the 84 xxh3 goldens.

A differential run against `origin/main` produces identical bytes for keyword, integer, negative integer, 1-, 2-, 3- and 4-byte UTF-8, mixed-width strings, unicode keywords, the empty string, and nested vectors. A 20,000-roll chain ends on the same seed on both engines and both host widths:

```
native before/after, TinyGo before/after -> final-seed = -2114297250817732603
```

The JVM cross-validation in `test/hash_test_jvm.clj` passes unchanged: 38 of 38 published vectors and 13 of 13 streaming cases. That file is an independent Clojure re-implementation rather than a caller of `xsofy.hash`, so it is left alone deliberately; mirroring this refactor into it would cost the independence that makes it worth having. Its `encode-salt` reaches the same bytes by a different route, `String.getBytes("UTF-8")` against our codepoint walk, and the two agree byte for byte on all four UTF-8 widths, a unicode keyword, a mixed-width string, the empty string, and a nested vector. That agreement is the one I most wanted to see: let-go iterates a string as codepoints and the JVM as UTF-16 chars, so an astral character such as U+1F600 is one unit on one side and a surrogate pair on the other.

Multibyte encoding was not covered before, and this change rewrites the encoder, so the cases above are now in the suite. Breaking the 3-byte lead byte previously passed all 328 tests: the string fixture was `"hi"`, and a comment in the test file recorded that the fixtures are "all printable ASCII; we deliberately don't go through utf8-bytes". Every mutation that changes output now fails, checked by reintroducing an error in the 2-, 3- and 4-byte lead bytes and in a 4-byte continuation byte.

### DIFF
--- a/xsofy/hash.lg
+++ b/xsofy/hash.lg
@@ -121,11 +121,31 @@
 (def ^:private salt-tag-str 3)
 (def ^:private salt-tag-vec 4)
 
+(defn- conj-int->8-be
+  "Append the low 64 bits of n to acc as 8 big-endian byte values, most
+   significant first.
+
+   The accumulator form is the one the hot path uses: every caller
+   immediately concatenates these 8 bytes onto a longer byte vector, so
+   returning a fresh vector meant allocating one per integer and then
+   copying it through `into`. Written as loop/recur rather than
+   (mapv f (range 8)) for the same reason — the higher-order form
+   allocates a range seq and a closure, and neither the IR optimizer nor
+   the Go lowerer can see through `mapv` to remove them."
+  [acc n]
+  (loop [i 0 acc acc]
+    (if (< i 8)
+      (recur (inc i) (conj acc (bit-and (u64-shr n (* (- 7 i) 8)) 0xFF)))
+      acc)))
+
 (defn- int->8-be
   "Big-endian byte encoding of the low 64 bits of n as a length-8 vector
-   of int byte values [0, 255]. Index 0 holds the most significant byte."
+   of int byte values [0, 255]. Index 0 holds the most significant byte.
+
+   Kept as a named function because the JVM cross-validation twin in
+   test/hash_test_jvm.clj mirrors it directly."
   [n]
-  (mapv #(bit-and (u64-shr n (* (- 7 %) 8)) 0xFF) (range 8)))
+  (conj-int->8-be [] n))
 
 (defn- codepoint->utf8
   "Encode a single Unicode codepoint as a vector of 1-4 int byte values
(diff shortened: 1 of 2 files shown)


mparrett:
Self-check:

No actionable findings in [[PR #194](https://github.com/nooga/xsofy/pull/194)](https://github.com/nooga/xsofy/pull/194). The accumulator encoding preserves ordering, recursive vectors, UTF-8 byte lengths, and macro-folded salt behavior.

Verified:

- Full suite: 330 tests, 2,864 assertions, zero failures
- JVM cross-validation: 38/38 vectors and 13/13 streaming cases
- Full application bytecode compilation succeeds
- GitHub pinned/latest checks pass; PR is mergeable

Vedict: Approve
Note: I did not independently reproduce the performance benchmarks.

nooga (approved):
Reviewed via automated multi-angle analysis (byte-level diffing of base vs. head, hand-traced induction proof for recursive/nested vectors, and a full test-suite run), since a "safe" perf refactor to a hash/salt encoder is exactly the kind of change that can silently break determinism if it's wrong.

**Verdict: approve.** Differentially ran `xsofy/hash.lg` on both `0e2a8b8` (base) and the PR head against a case exercising every term type at once (negative int, keyword, nested sub-vector, 4-codepoint-width UTF-8 string) — output is byte-identical, and `combine`/`xxh3-64` match too. Traced by hand that `encode-salt-into` is an inductive rewrite where `encode-salt-into(acc, s) == acc ++ encode-salt(s)` for every term type, including the "recursive vectors" claim in the PR body (`hash.lg:248-254` threads one accumulator through nested elements instead of building-and-`into`-ing per-element vectors — same order, verified not just asserted). `int->8-be`'s replacement (`hash.lg:120-135`) is still MSB-first, still 8 bytes, unchanged for negatives.

Test coverage is real, not just "hash still works": the new `hash_test.lg:189-212` cases pin literal byte arrays across all four UTF-8 widths plus a mixed-width string and empty string, closing a gap the old suite's own comment flagged (fixtures were "all printable ASCII"). Ran the suite myself on the PR branch: 330 tests / 2864 assertions, 0 failures — matches the PR body.

The allocation-reduction claim holds and isn't just moved allocations — the accumulator removes a per-append transient-wrapper allocation that `into` was paying on every codepoint, consistent with the measured 96→55 allocs/op.

Two non-blocking notes, not this PR:
1. `hash.lg:195-203` — the two `reduce` passes over a string (`utf8-len` then `conj-utf8`) each re-run `valid-codepoint!`, so multibyte codepoints get validated twice instead of once. Redundant-check cost only, not a correctness issue.
2. Unrelated to this diff: `xsofy/test/run.lg`'s trailing `(run-tests)` call takes zero args, which defaults to `(run-tests *ns*)` — only the `run` namespace itself, which defines no tests — so invoking it directly prints "Ran 0 tests" instead of the real suite. Worth a separate look since it means that entrypoint looks deceptively green locally with zero assertions.

Issue #187, opened 2026-08-20
Creature turn order follows map iteration order in process-creatures


mparrett:
`ai.lg:424 process-creatures` derives the order creatures act in straight from `(vals (:entities world))`, with no sort:

```clojure
(defn process-creatures [world attack-fn ranged-attack-fn]
  (let [creature-ids (->> (:entities world)
                          (vals)
                          (filter :ai)
                          (filter #(not= :player (:id %)))
                          (map :id)
                          (vec))]
    (reduce ...)))
```

Every creature action advances the `det` seed chain, so acting order is part of the determinism contract. Map iteration order is not something let-go promises — its `NewArrayMap` comment called the behavior "orderless by design" until nooga/let-go#764 replaced that with a specified contract. `find-target` at `ai.lg:150` already treats it that way, sorting on `(juxt dist-to (comp str :id))`, and its docstring says why:

> Ties on chebyshev distance break on `(str :id)` so equidistant candidates are
> picked deterministically from (seed, action-log).

`process-creatures` has the same dependency without the same guard.

## Why this has not bitten yet

Two things mask it, and neither is an invariant:

1. let-go's HAMT iteration order is content-determined, so it is stable across processes and across runs.
2. A real dungeon carries 19 to 34 entities, past the size where any ordered small-map representation would apply.

## What changes

nooga/let-go#764 gives maps of at most 8 entries insertion order, matching Clojure's `PersistentArrayMap`. Maps of 9 or more keep HAMT order, and `dissoc` never demotes, so real games are unaffected. I replayed seeds 1, 7, 42, 555, 9999 and 123456 through `dispatch/replay` with a 30-action script including a descend, under lg built at that PR and at its merge base: final seed, entity set, player position and turn count are bit-identical. The suite is 318 tests and 2825 assertions, passing identically under both.

#764 is safe for us today. The problem is that the only thing standing between us and a silent turn-order change is entity count staying above 8.

Repro, mirroring the `process-creatures` expression exactly:

```clojure
(def world
  {:entities (reduce (fn [m e] (assoc m (:id e) e))
                     {}
                     [{:id :player   :ai false}
                      {:id :rat-1    :ai true}
                      {:id :rat-2    :ai true}
                      {:id :goblin-1 :ai true}
                      {:id :bat-1    :ai true}])})
(->> (:entities world) (vals) (filter :ai)
     (filter #(not= :player (:id %))) (map :id) (vec))
```

```
[...]

PR #195, opened 2026-09-14
fix(build): pin let-go to v1.13.0, and fix the runner it silently zeroes

mparrett:
A bundle built with the pinned let-go does not start. It dies at namespace load:

```
error loading xsofy.input: ExecutionError: calling dev?
  caused by bit-shift-left expected Int
```

`dev?` runs when `xsofy.input` loads, so this is not a dev-mode-only path: the bundle never reaches the title screen.

## Why the pin was a SHA, and why it no longer needs to be

`.let-go-version` held the bare commit cf195a67, which is nooga/let-go#378 itself, taken on 2026-07-16 because no release carried the `#app` host mount yet. The file said so, and said what to do about it:

> Re-pin to a release tag once let-go ships #378.

v1.12.2 shipped #378 on 2026-07-20, so the SHA stopped being necessary four days after it was added. What this changes is not only that bookkeeping: the pin had also gone stale against the rest of the tree, and the failure above is the result.

This targets **v1.13.0**, released 2026-09-20, rather than v1.12.2. It carries #378 like v1.12.2 does, and also #859 and #862, which make `vm.Int` 64 bits on every host and lower a let-go `:int` to `int64`. Both are no-ops for this bundle — they widen 32-bit-int hosts (linux/386, TinyGo wasm), and stock Go wasm was already 64-bit — so the worlds existing seeds name are unchanged. What those seeds *play* is not — see “Worlds survive the bump; play does not” below.

## The runner fix is not optional

v1.13.0 ports `clojure.test` for real (nooga/let-go#863), which gives `run-tests` Clojure's actual arities: no arguments means the **current** namespace. `xsofy/test/run.lg` calls a bare `(run-tests)` from `xsofy.test.run`, a namespace holding no deftests. On v1.13.0 that runs zero tests and exits 0, so this lane would have gone green having executed nothing.

Same tree, same command, three runs:

```
v1.12.2                    Tests: 328  Pass: 2857  Fail: 0  Error: 0
v1.13.0                    Ran 0 tests containing 0 assertions      exit 0
v1.13.0 + run-all-tests    Ran 328 tests containing 2857 assertions, 0 failures
```

Selecting the namespaces by regex keeps the `:require` list as the source of truth, since `all-ns` only sees what has been loaded.

The release notes call out the opcode set and bundle format as breaking, but not this. Anything downstream using a bare `(run-tests)` as an aggregate runner has the same silent hole.

## Two warnings the bump surfaces

v1.13.0 warns when a namespace replaces a core referral, and the tree trips it twice:

```
WARNING: lazy-map already refers to: #'clojure.core/lazy-map in namespace: xsofy.check, being replaced by: #'xsofy.check/lazy-map
WARNING: peek already refers to: #'clojure.core/peek in namespace: xsofy.dag, being replaced by: #'xsofy.dag/peek
```

Both collisions predate this bump — core has carried `(def lazy-map map)` and `peek` since before the SHA being replaced — so what is new is the warning, not the shadowing.

They are fixed in opposite directions, because they are different situations.

`xsofy.check/lazy-map` is renamed to `lazy-map*`. It is private and file-local, and it is a different function from core's: core's is an alias for eager `map`, while check.lg's is a cons-based lazy version, one of the pair documented at check.lg:66 as existing because let-go's `map` and `mapcat` are eager and rose trees need laziness. Excluding the referral would assert a replacement this is not. Seven call sites, one file.

`xsofy.dag/peek` keeps its name, under `(:refer-clojure :exclude [peek])`. It is public, called from ten sites in dag_test.lg, and it is the domain term: reading the world at a commit-id in a namespace modelling a git-like DAG. Renaming it would churn a public API to quiet a warning, where excluding says what is true — in this namespace, `peek` means the DAG read.

The rule, for anyone hitting this on another bump: rename when the local is a different function that happens to share a name, exclude when the local is the intended meaning of that name. The assertion counts below are unchanged by either.

## Worlds survive the bump; play does not

The worlds a seed names are unchanged. What a seed plays is not.

`(world/make-world 40 25 12345)` on the current pin `cf195a67` and on v1.13.0 `369e2a69` returns the same `:seed`, the same 25 entities with byte-identical generated ids, and the same player position `[25 20]`. Worldgen is untouched.

What changes is the order `(keys (:entities w))` comes back in:

```
cf195a67 : :player :goblin-621174419 :rat-1400099564 :rat-39790946 …
v1.13.0  : :goblin-510791456 :rat-1355154828 :rat-1400099564 … :player 9th
```

`process-creatures` takes creature acting order straight from `(vals (:entities world))` — #187 — so the same seed plays out differently either side of this bump. Measured on the balance harness over 40 runs with the tree held fixed and the lg version the only variable, mean survival moves from 132.2 turns to 163.4.

The assertion counts below do not speak to this. The replay and determinism tests replay within one lg version, comparing a run against itself, so they pass either way.

Nothing here blocks the bump: the bundle is broken on the current pin and the fix is to move off it. But a replay code recorded before this lands is not guaranteed to reproduce after it, and #187 is what would make that guarantee hold. The evidence is on #187.

## Nothing shipped broken

`deploy-pages.yml` fires only on `v*` tags, and the last one, v0.0.2, predates the SHA pin. That release deployed with v1.11.1, which is why xsofy.quest currently serves a pre-#378 bundle mounting `#terminal` rather than `#app`. The live site is healthy.

The next tag would have been the first to carry the SHA pin, so this is a latent release blocker rather than an outage.

## Why the floor lane is green

The `pinned (deploy floor)` lane compiles each test namespace, runs the suite, runs the Layer-1 runtime smoke, and — since #69 — builds the WASM bundle through the same `build-wasm` action the deploy uses. What it never does is **boot** the bundle it just built. This failure is at namespace load, so every one of those steps passes on the old pin and the lane has no way to see it.

An earlier revision of this body, and of the `.let-go-version` comment, said the lane never builds the bundle. That was wrong, and the correction sharpens the point: the gap is not a missing build, it is a missing load. Issue #196 carries the same error and needs the same correction.

## Verification

On v1.13.0: 328 tests and 2857 assertions with the fixed runner, a full production-faithful bundle build, and the resulting bundle loaded headless under COI headers — `crossOriginIsolated` true, `#app` present, title screen, and the map screen after a keypress. The only console message is a 404 for `build-info.json`, which the local build path does not generate.

The same browser check against the old SHA pin fails, with the error above. CI installs the pin with `go install …@<version>`, which takes a tag and a SHA alike, so the resolution path is unchanged.


nooga (approved):
Reviewed via automated multi-angle analysis, including building `lg` from both the old pin (`cf195a67`) and `v1.13.0` and running the actual WASM build pipeline against both.

**Verdict: approve**, with one open question for the author and a pre-existing gap worth a follow-up.

**The re-pin itself is sound.** `v1.13.0` is a real, published tag (`go list -m -versions` confirms it against the module proxy) and is `cf195a67`'s descendant 290 commits later, including the #859/#862/#906 PRs cited. Verified those two are genuinely no-ops for this bundle: stock `GOOS=js GOARCH=wasm` already has 64-bit `int` (`PtrSize: 8` in Go's own `arch.go`), so they don't explain any crash even if the crash is real. `grep -rn "cf195a6"` across the repo hits only `.let-go-version` — nothing else needed updating.

**The "silently zeroes" half is confirmed real and correctly fixed, independent of the crash claim.** Not a `.let-go-version`-parsing issue (both CI extraction sites are format-agnostic between a SHA and a tag). The actual bug: at `cf195a67`, `run-tests` with zero args ran every already-loaded namespace with tests (`(doseq [[s bs] *registered-tests*] ...)`), which is what gave "328/2857." At `v1.13.0`, post the `clojure.test` port, zero-arg `run-tests` means *only the calling namespace* — `xsofy.test.run` defines no tests itself, so the old bare `(run-tests)` now silently reports 0 tests and exits 0. `run-all-tests #"xsofy\.test\..*"` (new in v1.13.0) is the correct semantic fix, not a workaround — restores coverage by construction rather than an explicit namespace list that could drift.

**One thing I couldn't verify:** I built the real WASM bundle at `cf195a67` via the actual `make wasm`/`build-wasm` action pipeline and booted `main.wasm` under Node with `wasm_exec.js`, with and without `?mode=dev` — no crash, no "error loading xsofy.input" in either case. Doesn't mean the claim is wrong (a JS-host or Worker-thread-specific path my single-threaded harness doesn't exercise could still trigger it), but it's currently unverified on my end. Would ask for the actual repro (browser + flags) so the root cause is nailed down, though it doesn't change whether re-pinning to a real tag is the right move regardless.

**Pre-existing gap, not introduced by this PR:** no automatic, non-label-gated, non-tag-gated CI lane actually boots the WASM bundle in a browser — `test.yml`'s WASM step only builds it. The only lanes that boot it (`tests/browser/smoke.mjs`) are gated behind a `v*` tag push or an opt-in PR label. So the exact failure mode this PR's headline bug describes (builds clean, crashes at runtime) would go unnoticed by default PR CI again in the future. Worth a follow-up issue, not a blocker here. Also: nothing asserts a minimum test/assertion count post-fix, and the runner doesn't fail the CI step on individual `is` failures — same class of silent-pass risk in a different shape.

Minor: the new `.let-go-version` comment says the floor lane "never builds or boots" the bundle — it does build it, just doesn't boot it. Small wording nit, not consequential.

Diff mechanics check out: `lazy-map` → `lazy-map*` rename in `xsofy/check.lg` is complete (def + all 6 call sites, private, no external callers possible), and `dag.lg`'s new `(:refer-clojure :exclude [peek])` is safe (no internal use of core `peek` that the exclusion could break). CI green on both required lanes.

mparrett:
## The repro

Chromium via Playwright, COI headers, no flags — plain `/`, no `?mode=dev`. Bundle built from this branch's tree with lg at `cf195a67`, through all four steps of the `build-wasm` action.

```
=== TERMINAL TEXT ===
[program exited]
xsofy v0.0.2

=== CONSOLE ===
[log] error loading %s: %v
%!(EXTRA string=xsofy.input, *vm.ExecutionError=ExecutionError: calling dev?
	caused by bit-shift-left expected Int)
```

The boot probe times out after 60s waiting for the title screen; `#app` mounts and `crossOriginIsolated` is true, so the host comes up and the guest dies at namespace load.

Two details that make this easy to miss, and may be why your harness came back clean. It arrives as `console.log`, not `console.error` or stderr, so a check that watches the error channel sees nothing. And the format string is botched — `error loading %s: %v` is printed literally with the arguments trailing in `%!(EXTRA ...)` — so a grep for the assembled message finds nothing either.

`dev?` is `(or (= (js/url-param "mode") "dev") (os/getenv "XSOFY_DEV"))` (`input.lg:5-12`). Under Node with `wasm_exec.js` there is no `location`, so the `js/url-param` path is not the same one a browser takes. I have not pinned down which of the two calls shifts a non-Int at that pin, so treat the mechanism as unconfirmed — the failure and its trigger are not.

## You were right about the floor lane

It does build the bundle. `test.yml` has run the `build-wasm` action since #69, so the lane builds exactly what the deploy builds; what it never does is load the result. Corrected in `.let-go-version` and in the PR body.

The correction sharpens the point rather than blunting it: the gap is not a missing build, it is a missing boot, and a build-only lane cannot catch a namespace-load failure by construction.

#196 carries the same wrong claim in its title and body and needs the same fix; it already tracks the CI gap you flagged, including the assertion-count floor.

PR #198, opened 2026-09-20
fix(console): correct the bench help line to µs

mparrett:
The console's `(bench)` returns `perf/summarize`'s map verbatim, and that map has reported `:unit :us` since the harness moved to microseconds. The help line above it was never updated, so the panel advertises ms over an avg of 8360:

```
│   (bench)    quick engine micro-benchmark (per-turn ms)            │
│ » (bench)                                                          │
│ {:n 60                                                             │
│  :unit :us                                                         │
│  :avg 8360                                                         │
```

Help text only. `bench/native.lg` is unaffected — it converts µs to ms itself for its table, so its `avg-ms` headers are right.

Checked by running the console's exact expression under lg v1.12.2: `{:name :wait, :n 60, :avg 8463, :p95 10680, :max 15002, :p50 8426, :unit :us}`


nooga (approved):
One-line doc fix, matches the actual `perf/summarize` unit (`:unit :us`, confirmed in the console output pasted in the PR body). Nothing else touched.

Issue #82, opened 2026-06-22
TUI: non-blocking modal/menu input


mparrett:
**Layer:** xsofy TUI (`xsofy/main.lg`, `xsofy/xsofy/ui.lg`)
**Parent:** nooga/xsofy#84 (Player interaction ergonomics). Closes the remaining semantic work in #56.

## Problem

The death screen runs on the poll-based `xsofy.ui/run-loop` (cosmetic clock pulses while waiting for a key). Every *other* modal screen still calls blocking `term/read-key` directly inside `game-loop` — so once a menu is open the loop is frozen: no clock ticks, no resize handling, and under WASM the worker can't pump cosmetic animation. The menus work but they're a different, more brittle input path than the main loop.

Blocking sites in `xsofy/main.lg` `game-loop` (current):
- `:messages` (~382), `:rune-codex` (~392), `:help` (~399), `:inventory` (~406), `:inscribe` (~415), `:quick-menu` (~424)
- `:confirm-hazard` (~505), `:confirm-quit` (~522), and the WASM death-screen restart confirm (~490)

## Repro

1. Open any menu (inventory `i`, help `?`, message log).
2. Resize the terminal (or rotate the phone) while the menu is open → layout doesn't re-fit until the menu closes.
3. Under WASM, note the menu has no cosmetic tick available; only the death screen animates.

## Fix shape

Convert each blocking modal path to a nested `ui/run-loop` with the same `:sources [(input-source …) (clock-source …)]` shape the death screen already uses. The existing `*-input` reducers (`inventory-input`, `quick-menu-input`, `messages-input`, `inscribe-input`) become the `:step` scan; `render/*-screen` becomes `:render`; exit via `(reduced …)`. Keep determinism: the input source must stay replay-safe (same key→event mapping as the main loop), so recorded runs still reproduce.

## Acceptance

- Every modal listed above runs through `run-loop`; no direct `term/read-key` left in `game-loop` branches.
- Resize while a menu is open re-fits the menu (native + phone browser).
- Replay determinism unchanged: existing replay fixtures pass; a menu-heavy replay round-trips.
- WASM: menus don't stall the worker; cosmetic clock available to modal renders that want it.

## Dependencies

- Builds on #56 (`run-loop`, `input-source`, `clock-source` already landed).
- Pairs with the resize/fit child — both want the loop live during modal display.

mparrett:
[...]

PR #200, opened 2026-09-20
tui: run modal screens on the poll loop instead of blocking reads

mparrett:
Closes the semantic half of #82.

Every modal in `game-loop` blocked on `term/read-key`. While a menu was open the loop was frozen: no clock ticks, no resize handling, and under wasm nothing to pump the worker. The death screen and the console already ran on `ui/run-loop`, so the menus were a second and more brittle input path.

## What changed

All nine blocking sites now go through one of two helpers, and `main.lg` has no remaining `term/read-key` call.

`modal-loop` covers the six screen branches — `:messages`, `:rune-codex`, `:help`, `:inventory`, `:inscribe`, `:quick-menu`. It takes each screen's existing key reducer unchanged and hands it the same raw key the blocking path did, so the key-to-world mapping is identical and replays are unaffected.

`prompt-key!` covers the three y/n confirms — `:confirm-hazard`, `:confirm-quit`, and the wasm death-restart — which return a key rather than a world.

`read-dismiss-key!` is replaced by `dismiss-input`.

## Where the conversion is not mechanical

**`modal-loop` exits when `:screen` changes, not when it clears.** `inventory-input` and `inscribe-input` hand off to each other by reassigning `:screen`. Exiting only on a cleared screen would keep painting the inventory while the world had moved to inscribe; exiting on any change lets the caller's `case` dispatch, which is what the blocking code did.

**`prompt-key!` does not repaint the game screen every frame.** The confirms previously painted `render-full` once and then blocked. Painting it per frame would be a cost the blocking version never paid, so the repaint is gated on the terminal size changing. The prompt line itself is painted every frame.

`dismiss-input` keeps `read-dismiss-key!`'s BEL filter. The shell still sends a lone BEL to nudge a blocked read on browser resize, and a dismiss-on-any-key screen would otherwise take it as the dismiss key.

## Verification

Suite: 328 tests, 2857 assertions, 0 failures — unchanged from `main`.

Native, under tmux, resizing with each screen open:

| | `main` | this branch |
|---|---|---|
| help box after pane 100 to 130 cols | stays 104 | re-fits to 134 |
| inventory after pane 130 to 90 cols | — | re-fits to 94 |

The `main` column is the control: the same steps against `main` leave the box at its old width, so the check distinguishes rather than passing on both.

Also exercised: `?` opens help and any key dismisses it, `i` opens the inventory and Esc closes it, Esc opens the quit confirm, which survives a resize with the game screen re-centring behind it, and `n` returns to play.

Not covered here: menus under wasm. The change is what makes them pumpable, but I have not driven the browser menus, and someone should before this is trusted on a phone.

## Left for the follow-up

#82 also asks that a menu-heavy replay round-trips. The reducers and the key mapping are untouched, so nothing should have moved, but there is no fixture that opens a menu mid-replay to prove it.


main.lg
```diff
@@ -358,20 +358,70 @@
...
+       :step     (fn [st ev]
+                   (let [[kind payload] ev]
+                     (if (and (= kind :action) (not= payload bel))
+                       (reduced (if (= payload :eof) nil payload))
+                       st)))
+       :render   (fn [_]
+                   (let [size (term/size)]
+                     (when (not= size @painted)
+                       (reset! painted size)
+                       (render/render-full world)))
```
mparrett: [P2] Preserve the death screen behind this prompt. In the WASM `:death` branch, Esc now calls `prompt-key!` with the dead world. `painted` starts nil, so the first render always calls `render/render-full world`, which clears the tombstone and animated runes and replaces them with the gameplay map while `Start a new game?` waits for input. The previous branch only drew the prompt over the death screen. A resize in this prompt has the same effect; it needs the death renderer (or a caller-specific background render) instead of `render-full`.

main.lg
```diff
@@ -358,20 +358,70 @@
...
+  (let [from (:screen world)]
+    (ui/run-loop
+      {:sources  [(ui/input-source ui/raw-key->event) (ui/clock-source 0)]
+       :state    world
+       :step     (fn [w ev]
+                   (let [[kind payload] ev]
+                     (if (= kind :action)
+                       ;; raw-key->event maps EOF to :eof; the reducers expect nil.
+                       (let [w' (input w (if (= payload :eof) nil payload))]
+                         (if (= (:screen w') from) w' (reduced w')))
```
mparrett: [P2] Keep keys that follow a screen transition in the same poll batch. `ui/input-source` drains up to 16 pending keys before `:step` runs, but `(reduced w')` stops `apply-events` and the rest of that already-drained batch is discarded. For example, if inventory receives `a`, Esc, `j` in one frame, `a` selects, Esc closes, and `j` never reaches the play loop; the old blocking reads left `j` queued. This also affects quick menu handoffs. Poll one key at a time for these nested loops or carry unconsumed events into the next screen.

main.lg
```diff
@@ -409,55 +459,42 @@
...
 
          :quick-menu
-         (do (render/render-quick-menu world)
-             (let [k (term/read-key)
-                   w (quick-menu-input world k)]
-               (if (:screen w)
-                 (recur w w prev-size)
-                 (do (render/render-full w)
-                     (recur w w (term/size))))))
+         (let [w (modal-loop world render/render-quick-menu quick-menu-input)]
```
mparrett: [P2] Clear the old quick-menu rectangle when the terminal size changes. `render-quick-menu` draws only its newly centered box and clears only that box's inner rectangle. This loop invokes it at the new coordinates after a resize without repainting the game or erasing the previous box, so a larger native terminal leaves remnants of the old menu on screen until the menu closes. Track size and repaint the background before drawing the moved box.

mparrett (commented):
Review findings on the modal poll-loop conversion. Local suite verification was unavailable: the installed `lg` binary is at 369e2a6, while this repo pins cf195a6, and it reported 0 tests in the isolated PR tree.

nooga (changes requested):
Reviewed via automated multi-angle analysis (line-by-line trace, direct reproduction against the actual `ui/run-loop`/`input-source` machinery rather than just reading it, and a local test run).

**Requesting changes: two real bugs, one of them in exactly the handoff this PR calls out by name.**

1. **Keystroke drop on `:screen` handoff mid-batch.** `ui/input-source` (`xsofy/ui.lg:206-219`) drains all currently-pending keys into one batch per poll tick (up to 16), and `ui/apply-events` (`xsofy/ui.lg:267-276`) stops folding the batch the instant a reducer returns `reduced`. So if two keystrokes are already buffered when a `modal-loop` reducer changes `:screen` on the first one, the second key has already been consumed off the terminal but is never delivered to any reducer — it's silently gone, and the next screen's `modal-loop` starts a fresh `input-source` with no memory of it. I reproduced this directly (not a mock) using `input-source`'s test-injectable 3-arity, queuing "a" then "b" against a reducer that hands `:inventory` → `:inscribe` on "a" — the exact scenario the PR body highlights: "b" is read off the terminal but never reaches a reducer. The old blocking code never had this failure mode, since it read one key at a time and a second buffered key would always be picked up by the next screen's read, in order. Realistic triggers: fast typing, key-repeat while holding a key (~30-50ms repeat intervals land inside the 30ms poll window), or pasting into the inscribe prompt right after selecting an inventory item. This is a real regression in "the key-to-world mapping is identical" claim, not a pre-existing issue introduced elsewhere — worth either an explicit accepted-risk note or re-queuing the undelivered tail of the batch into the next loop instead of discarding it. (To be fair: the drop-on-`reduced`-mid-batch shape isn't new — `play.lg:77-88` and `title.lg:181-200` already have it at 2 sites — but this PR extends it from 2 to 8 call sites.)

2. **`prompt-key!` never repaints off a TTY.** `main.lg:403-424` gates the repaint on `(not= (term/size) @painted)` with `painted` initialized to `nil`. On a real TTY this is fine — first read is a non-nil `[w h]`, not-equal to `nil`, so it paints once on entry then again only on genuine resizes. But `term/size` (`pkg/rt/term.go:417-436`) explicitly returns `nil` when stdout isn't a TTY — confirmed with `lg -e '(term/size)'` off a TTY. In that case `size` is `nil` every frame, `@painted` starts at `nil`, `(not= nil nil)` is `false`, and the gate never fires — `render-full` is never called for the whole prompt lifetime, off a TTY. The old code painted unconditionally once per entry regardless of TTY-ness, so this is a real (if narrow, native-non-TTY-only) regression. Cheap fix: seed `painted` with a sentinel that can't equal a real size (e.g. `::none`) instead of `nil`.

Gap, not a blocker on its own but worth calling out: no tests were added for `dismiss-input`/`modal-loop`/`prompt-key!` (diff touches only `main.lg`). The codebase already has the right pattern for this — `ui/input-source`'s 3-arity and `ui/clock-source`'s 2-arity are explicitly documented as test-injectable — it's just that these three new functions live in `main.lg`, which can't be required in isolation since it invokes `-main` at load time. A thin seam that lets them be exercised headlessly would have caught bug #1 directly.

Confirmed correct, no action needed: `modal-loop`'s screen-change exit condition matches its description exactly (needed for the inventory↔inscribe handoff to repaint the right screen); `dismiss-input`'s BEL-filtering reproduces `read-dismiss-key!`'s old semantics exactly, including the real-Ctrl-G-vs-resize-nudge ambiguity and EOF handling; all 9 former blocking call sites are converted (`grep -n term/read-key main.lg` → zero matches, 6 `modal-loop` + 3 `prompt-key!` sites confirmed); no busy-wait or resource leak (`pause!`/`timeout`-paced, pre-existing pattern); replay-log integrity holds, though for a more specific reason than "unaffected" — modal handlers mutate the world outside `:action-log` by design, so replays never drive `modal-loop` at all today, which also means bug #1 can't desync a replay, it just loses input during live play. Local suite (workaround needed: `xsofy/test/run.lg`'s `(run-tests)` call is 0-arg and only runs its own empty namespace — pre-existing, unrelated to this PR, but worth a separate look since it makes `make test` look green with zero assertions) came back at 328 tests / 2857 assertions / 0 failures, matching the PR body.

mparrett:
All four findings reproduced before fixing, and the branch is rebased onto main now that #191 and #194 have landed. Suite is 336 tests and 2876 assertions, 0 failures.

## Keys dropped on a `:screen` handoff

Confirmed, and it is the worse of the two. `input-source` drains every pending key into one batch and `apply-events` stops folding at the first `reduced` (`ui.lg:273-276`), so a key already read off the terminal when a modal exits reaches no reducer at all.

Fixed by reading one key per tick in these nested loops, via a new `single-key-source`, rather than by re-queuing the tail. Unread keys then stay in the terminal buffer, which is what the blocking `read-key` calls these loops replaced did — the next screen picks them up in order without anything having to hand them over.

The pre-existing instances in `play.lg` and `title.lg` are left alone; they are a separate change against a loop this PR does not touch.

## `prompt-key!` off a TTY

Confirmed: `term/size` returns nil off a TTY, `painted` started at nil, and `(not= nil nil)` gated the first paint away for the prompt's whole lifetime. The sentinel is now `:unpainted`, which no size can equal.

## The death screen behind the restart prompt

Confirmed, and this one was a straight regression: the wasm restart confirm repainted the dead world with `render-full`, replacing the tombstone and runes with the map. `prompt-key!` now takes the background renderer from its caller, and that site passes nil, so the prompt draws over the death screen exactly as it did before.

## Quick-menu remnants after a resize

Confirmed. `render-quick-menu` centres a small box and clears only that box, so a resize drew it at the new centre with the old one still on screen. `modal-loop` takes an optional `:background!`, called on entry and on each size change rather than per frame, and the quick-menu site passes `render-full`.

Checked under tmux at 120 then 170 columns, and falsified by removing the option again: two box fragments on screen without it, one with it.

## The test gap

Taken. The three helpers move to `xsofy.ui`, which is what makes them reachable — `main.lg` calls `-main` at load and cannot be required from a test. Six tests now drive them through the injectable source and size seams, including the handoff case, the nil-size case, and the resize repaint count.

One of them asserts the *batching* behaviour that caused the first bug: if `input-source` ever stops draining, that test fails and the reason `single-key-source` exists stops being load-bearing.

## The `run-tests` note

Both of you flagged it, and it is already fixed in #195, which repins to v1.13.0 and switches the runner to `run-all-tests`. Worth knowing that the 0-arg behaviour is v1.13.0's — on the current pin a bare `(run-tests)` runs everything loaded, so the suite is real today and silently empties on the bump. Filed upstream as nooga/let-go#929.

